PR #31, opened 2020-10-30
add namesapces lookup by groups

### LINKED ISSUE #30: The namespaces list is returned empty when the tenant owner is a group
Currently the namespaces list is returned only when the tenant owner is a user.

## Current behaviour:

Having a tenant with a group as owner:

```
apiVersion: capsule.clastix.io/v1alpha1
kind: Tenant
metadata:
  name: spectre
spec:
  owner:
    kind: Group
    name: spectre
...
```

and a user being part of the group:

```
user info:
{
  "groups": [
    "capsule.clastix.io",
    "spectre"
  ],
  "sub": "63170aff-70ac-4f6a-8c6a-4691c97aaa65",
  "email_verified": false,
  "preferred_username": "adriano"
}

```

As admin:

```
$ kubectl get tenant spectre

NAME      NAMESPACE QUOTA   NAMESPACE COUNT   OWNER NAME   OWNER KIND   AGE
spectre   100               2                 spectre      Group        1d

```

As user being part of tenant owner group

`$ kubectl get ns`

From logs:

```
2020-10-27T22:10:02.731Z        LEVEL(-2)       namespace_filter        Decorating request for Namespace filtering
2020-10-27T22:10:02.731Z        LEVEL(-4)       namespace_filter        Getting user from request       {"username": "adriano", "groups": ["capsule.clastix.io", "spectre"]}
2020-10-27T22:10:02.731Z        LEVEL(-4)       namespace_filter        labelSelector added     {"selector": "dontexistsignoreme"}
2020-10-27T22:10:02.731Z        LEVEL(-4)       namespace_filter        updating RawQuery       {"query": "labelSelector=dontexistsignoreme&limit=500"}
2020-10-27T22:10:02.731Z        LEVEL(-4)       namespace_filter        Updating the token      {"token": "eyJhbGciOiJSUzI1NiIsImtpZC

```

## Desired behaviour:

As user being part of tenant owner group

```
$ kubectl get ns
NAME                 STATUS   AGE
spectre-marketing    Active   5s
spectre-production   Active   103m
```

bsctl:
Fixes #30. @prometherion please review.

### DIFF
--- a/internal/webserver/webserver.go
+++ b/internal/webserver/webserver.go
@@ -275,7 +275,6 @@ func (n kubeFilter) getTenantsForOwner(ownerKind string, ownerName string) (tena
 }
 
 func (n kubeFilter) getLabelSelectorForOwner(username string, groups []string) (labels.Selector, error) {
-	var r *labels.Requirement
 	capsuleLabel, err := capsulev1alpha1.GetTypeLabel(&capsulev1alpha1.Tenant{})
 	if err != nil {
 		return nil, fmt.Errorf("cannot get Capsule Tenant label: %s", err.Error())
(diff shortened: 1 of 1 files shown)


internal/webserver/webserver.go
```diff
@@ -252,35 +252,64 @@ func (n kubeFilter) getOwnedNamespacesForUser(username string) (res []string, er
...
-	if len(ownedNamespaces) > 0 {
-		r, err = labels.NewRequirement(capsuleLabel, selection.In, ownedNamespaces)
-	} else {
-		r, err = labels.NewRequirement("dontexistsignoreme", selection.Exists, []string{})
+	if len(ownedTenants) > 0 {
+		r, err = labels.NewRequirement(capsuleLabel, selection.In, ownedTenants)
+		if err != nil {
+			return nil, fmt.Errorf("cannot parse Tenant selector: %s", err.Error())
+		}
+		return labels.NewSelector().Add(*r), nil
```
prometherion, line 293: What happens if a Tenant Owner got a Tenant assigned by Name and also a subset of Tenant assigned by the group?

According to this implementation, we would miss the latter ones.

I would suggest refactoring this a little bit, trying to avoid as much as possible the if/else and using a final return statement.

prometherion (changes requested):
We should cover also the case of a User owning Tenant via User reference and Group as well.

internal/webserver/webserver.go
bsctl: Now covered the mentioned case too.

prometherion (approved):
LGTM

PR #33, opened 2020-11-04
Fix issues/32

bsctl:
@prometherion please review


prometherion:
It seems there are some conflicts, probably not starting from `master`?

PR #38, opened 2020-12-15
support for /api/v1/nodes filtering

### LINKED ISSUE #21: Access nodes list for tenant owner
# Describe the feature
As evolution for this project, we need to add an ability to also list nodes by tenant labels.
When user asks for a tenant, he is able to provision nodes for his tenant in different clouds using different pre-defined node flavours (BYOD - Bring Your Own Device). We will add specific labels to nodes, showing in which cloud this node is, what interfaces it has (backend, frontend) and other specific labels. I would be useful, the tenant owner is able to list his own nodes only.

1. the Cluster Admin creates the Tenant adding the nodeSelector field and the .metadata.annotations pair `capsule.clastix.io/enable-node-listing: "true"`
2. the Tenant Owner with support of the Cluster Admin (getting joining token from control plane requires admin permissions) adds the nodes marked with the label `capsule.clastix.io/tenant=${tenant_name}` labels provided
3. if the Tenant Owner is NOT entitled to list their nodes, we don't intercept and let Kubernetes handle it with the RBAC error.
4. with the `capsule-ns-filter` we're going to add add a new endpoint for nodes, i.e. `/api/v1/nodes` retrieving the available Tenant resources for the user and building the labelSelector as we're doing with Namespaces.

# Expected behavior
As tenant owner of a tenant with `capsule.clastix.io/enable-node-listing: "true"`, I should be able to list my own nodes only, e.g. `kubectl get nodes`.

Make sure the cluster admin only sets the `capsule.clastix.io/enable-node-listing: "true"` and does not give RBAC permissions to list nodes at cluster. If for some reason, the tenant owner is able to bypass the filter and directly asks the `kube-apiserver`, he gets the entire list of cluster nodes.

MaxFedotov:
closes #21

### DIFF
--- a/internal/request/http.go
+++ b/internal/request/http.go
@@ -25,6 +25,12 @@ func NewHttp(request *h.Request, usernameClaimField string) Request {
 	return &http{Request: request, usernameClaimField: usernameClaimField}
 }
 
+func (h http) IsNodeListing() (ok bool) {
+	ok = h.URL.Path == "/api/v1/nodes"
+	ok = (h.Method == "GET" || h.isWatchEndpoint()) && ok
+	return
+}
+
 func (h http) IsNamespaceListing() (ok bool) {
 	ok = h.URL.Path == "/api/v1/namespaces"
 	ok = (h.Method == "GET" || h.isWatchEndpoint()) && ok
--- a/internal/request/request.go
+++ b/internal/request/request.go
@@ -3,4 +3,5 @@ package request
 type Request interface {
 	GetUserAndGroups() (string, []string, error)
 	IsNamespaceListing() (ok bool)
+	IsNodeListing() (ok bool)
 }
--- a/internal/webserver/webserver.go
+++ b/internal/webserver/webserver.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httputil"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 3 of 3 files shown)


prometherion:
I'm trying to review this but it seems some functions have been moved up and down and I'm struggling to review it.

I promise I'll be back with this 🙏🏻

MaxFedotov:
@prometherion yep, i'm sorry for this. Had to do it in order to reuse existing code

bsctl:
@MaxFedotov I'll do a try on our setup.

bsctl:
@MaxFedotov from early tests, it works fine.

Noticed the only nodes listing is permitted

```
$ kubectl get nodes 
NAME                STATUS   ROLES    AGE   VERSION
k8s-caas-worker01   Ready    worker   96d   v1.19.1
k8s-caas-worker02   Ready    worker   96d   v1.19.1
k8s-caas-worker03   Ready    worker   96d   v1.19.1
k8s-caas-worker04   Ready    worker   96d   v1.19.1
k8s-caas-worker05   Ready    worker   96d   v1.19.1
k8s-caas-worker06   Ready    worker   96d   v1.19.1

$ kubectl get node k8s-caas-worker01
Error from server (Forbidden): nodes "k8s-caas-worker01" is forbidden: User "alice" cannot get resource "nodes" in API group "" at the cluster scope
```

This is correct according to the initial feature request. What about to give the user the option to get also the node item? For example inspecting its status and monitoring? In some scenarios, e.g. dedicated nodes, this may have sort of sense. @MaxFedotov @prometherion what do you think?

prometherion:
Well, it makes perfectly sense but require a huge refactoring regarding the http server route handler.

What we could do is to open another issue to get it, I can take care of it.

prometherion (approved):
Approving, although requiring a little bit of polishing but I'll address in the future.

PR #49, opened 2021-01-07
correct default image tag to match published image

d-m:
The image tag defaults to `appVersion` which is currently `0.0.5`. However, the image is actually tagged `v0.0.5`.

### DIFF
--- a/charts/capsule-proxy/templates/deployment.yaml
+++ b/charts/capsule-proxy/templates/deployment.yaml
@@ -38,7 +38,7 @@ spec:
         - name: {{ .Chart.Name }}
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default (printf "v%s" .Chart.AppVersion) }}"
+          image: {{ include "capsule-proxy.fullyQualifiedDockerImage" . }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command:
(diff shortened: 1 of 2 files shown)


d-m:
Fixes #50

charts/capsule-proxy/templates/deployment.yaml
```diff
@@ -38,7 +38,7 @@ spec:
         - name: {{ .Chart.Name }}
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default (printf "v%s" .Chart.AppVersion) }}"
```
prometherion, line 41: Maybe we could use something like this?

https://github.com/clastix/capsule/blob/a2109b74ef24ab5afe25379f3fb2ef43e916766c/charts/capsule/templates/_helpers.tpl#L64-L69

prometherion (commented):
Thanks for catching this, @d-m: really appreciated it!

Just suggested a small change, I'd like to hear feedback from you so we can plan the release!

charts/capsule-proxy/templates/deployment.yaml
d-m: Sounds good. I'll update the PR.

prometherion (approved):

PR #62, opened 2021-03-14
Move from http library to Gorilla Mux

### LINKED ISSUE #54: Migrate to a better HTTP router
Actually, the HTTP server is using the standard library `http` that provides a basic router that doesn't allow to perform path variables matching.

What we have to implement is something as this:

```
r := mux.NewRouter()
r.HandleFunc("/api/v1/nodes", NodeListHandler)
r.HandleFunc("/api/v1/nodes/{name}", NodeHandler)
```

Implementing this enhancement gives us the chance to address and close #40, but last and not least, it's going to give us more flexibility on addressing future issues, as well as dealing in a better way with middleware and request filtering.

Along with this, we can also implement in a better way the middleware chains we're using, as:

https://github.com/clastix/capsule-proxy/blob/c34b1aa822c3e0fb958d76bab5d3b4466e3ed533/internal/webserver/webserver.go#L234-L235

I like [gorillamux](https://github.com/gorilla/mux) as well as [gin-gonic](https://github.com/gin-gonic/gin): probably the first choice is more reliable since it's standard-library compatible.

davideimola:
The PR has been created for changing the library for HTTP requests.

The work has been made in pair programming with @ludusrusso!
Thx @prometherion for your help! 

Closes #54

### DIFF
--- a/internal/webserver/webserver.go
+++ b/internal/webserver/webserver.go
@@ -13,6 +13,7 @@ import (
 
 	capsulev1alpha1 "github.com/clastix/capsule/api/v1alpha1"
 	"github.com/clastix/capsule/pkg/utils"
+	"github.com/gorilla/mux"
 	v1 "k8s.io/api/authentication/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
@@ -230,19 +231,20 @@ func (n kubeFilter) isNodeListing(fn http.HandlerFunc) http.HandlerFunc {
 }
 
 func (n kubeFilter) Start(ctx context.Context) error {
-	http.HandleFunc("/api/v1/namespaces", func(writer http.ResponseWriter, request *http.Request) {
+	r := mux.NewRouter()
+	r.HandleFunc("/api/v1/namespaces", func(writer http.ResponseWriter, request *http.Request) {
 		n.isNamespaceListing(n.checkJwt(n.namespacesHandler))(writer, request)
 		n.reverseProxyFunc(writer, request)
 	})
-	http.HandleFunc("/api/v1/nodes", func(writer http.ResponseWriter, request *http.Request) {
+	r.HandleFunc("/api/v1/nodes", func(writer http.ResponseWriter, request *http.Request) {
 		n.isNodeListing(n.checkJwt(n.nodesHandler))(writer, request)
 		n.reverseProxyFunc(writer, request)
 	})
-	http.HandleFunc("/_healthz", func(writer http.ResponseWriter, request *http.Request) {
+	r.HandleFunc("/_healthz", func(writer http.ResponseWriter, request *http.Request) {
 		writer.WriteHeader(200)
 		_, _ = writer.Write([]byte("ok"))
 	})
-	http.HandleFunc("/", func(writer http.ResponseWriter, request *http.Request) {
+	r.PathPrefix("/").HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
 		n.reverseProxyFunc(writer, request)
 	})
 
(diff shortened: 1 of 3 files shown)


internal/webserver/webserver.go
```diff
@@ -21,6 +21,7 @@ import (
 
 	"github.com/clastix/capsule-proxy/internal/options"
 	req "github.com/clastix/capsule-proxy/internal/request"
+	"github.com/gorilla/mux"
```
prometherion, line 24: Import statements are not in the correct order.

Third packages as `github.com/gorilla/mux` must be in the second section, resulting as this:

```go
import (
        // this packages are from the standard library, they are first class citizens
	"context"
	"crypto/tls"
	"encoding/json"
	"fmt"
	"net/http"
	"net/http/httputil"
	"strconv"
	"strings"
	"time"

        // third party packages, in the second block
	capsulev1alpha1 "github.com/clastix/capsule/api/v1alpha1"
	"github.com/clastix/capsule/pkg/utils"
	"github.com/gorilla/mux"
	v1 "k8s.io/api/authentication/v1"
	"k8s.io/apimachinery/pkg/labels"
	"k8s.io/apimachinery/pkg/selection"
	ctrl "sigs.k8s.io/controller-runtime"
	"sigs.k8s.io/controller-runtime/pkg/client"

        // third block, internal packages
	"github.com/clastix/capsule-proxy/internal/options"
	req "github.com/clastix/capsule-proxy/internal/request"
)
```

JFI you may notice `github.com/clastix/capsule/pkg/utils` in the second block and an eyebrow could rise: that's ok since it's from the `capsule` repository, it's not `capsule-proxy`, so it has to be treated as a third package since it's also imported using the Go modules.

prometherion (changes requested):
Minor change regarding import statement: we don't have yet the linting check, but since we're in pursuit of dragons, let's get this done! :)

@bsctl may I ask you a functional review for this? Unfortunately, we don't have yet e2e testing here, we need your QA super powers.

internal/webserver/webserver.go
```diff
@@ -230,19 +231,20 @@ func (n kubeFilter) isNodeListing(fn http.HandlerFunc) http.HandlerFunc {
 }
 
 func (n kubeFilter) Start(ctx context.Context) error {
-	http.HandleFunc("/api/v1/namespaces", func(writer http.ResponseWriter, request *http.Request) {
+	r := mux.NewRouter()
+	r.PathPrefix("/api/v1/namespaces").HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
```
prometherion, line 235: I was thinking about the need for `PathPrefix` here, as I suggested to you, but I guess we don't need it: we are solving the missing listing capabilities at the cluster lever for the Tenant Owner, and that's good, but we don't have to decorate the request for a specific resource (`/api/v1/namespaces/{namespace}`) since the Tenant Owner will be granted by the `capsule` operator using the proper RBAC.

I think we can safely switch to `r.HandleFunc`.

prometherion (changes requested):

internal/webserver/webserver.go
davideimola: Ok :+1:

bsctl:
@davideimola @ludusrusso thanks for the great work.
@prometherion I would suggest to merge this PR when #3 is in place since it could be a breaking change from the code base point of view. We have the current code base in production and an extensive manual QA keeps time.

prometherion (approved):

prometherion:
> @prometherion I would suggest to merge this PR when #3 is in place since it could be a breaking change from the code base point of view. We have the current code base in production and an extensive manual QA keeps time.

My idea was to develop on the main branch for the upcoming major release and, in case of hotfix required for the old version, backport it in a separate trunk.

Anyway I couldn't agree more on putting higher priority on the e2e suite, going to take it.

bsctl (changes requested):
@davideimola the proxy is not working for regular paths that do not need to be decorated, eg. `/api` `/api/v1/pods` `/version`. Paths that should be decorated like `/api/v1/namespaces` and `/api/v1/nodes` work fine. The only other path working is `/` so I suspect the Gorilla Mux is managing the catch-all handler differently than the regular Go router.

davideimola:
Hi @bsctl !
Sorry, it was my fault! Now it might be working fine!

prometherion (approved):
Performed also an integration test on my machine: LGTM!

bsctl (approved):
lgtm @davideimola @ludusrusso

prometherion:
@davideimola please, could you rebase over `master` since #69 has been merged and we got automated tests?

davideimola:
@prometherion done! Hope the test goes green :smile:

PR #66, opened 2021-03-14
Add an option to pass the k8s api token from command line to simplify local debuggin and add docs on how to run the proxy locally

ludusrusso:
Tnx to @davideimola and @prometherion!


ludusrusso:
Now you can run the proxy with `--k8s-api-token=<K8S API TOKEN>`!

@prometherion what do you think?

MaxFedotov:
Guys, i don't think we had to add additional fields to structs only for debugging purposes, as there are already a lot of ways for remote debugging in kubernetes, for example https://dchaykin.medium.com/debugging-a-go-service-in-kubernetes-f1dd48865e1d

ludusrusso:
I think this is useful to simplify local debugging. It is true that you can debug in cluster, but it requires some effort to do it!
With this flag, you can easily run the proxy using your preferred ide and with debugger active with very small effort. 

I worked on #62 with @davideimola and we were able of testing locally only thanks to @prometherion

prometherion:
I guess the enhancement proposed by @ludusrusso can be merged, although I wouldn't sponsor it for debug purpose but rather for the running mode _out of the cluster_.

I saw this pattern pretty frequently, and `capsule-proxy` right now cannot run outside of a Pod: since we're using the `controller-runtime` configuration auto-discovery feature and we don't provide any limitation on how to run, we could use this feature as a requirement for the new running mode.

Said so, we need some refactoring, since we have to put a validation to understand if `capsule-proxy` is running out of the cluster and, if it's the mode, check if the optional flag `--k8s-api-token` (although I would suggest using `--bearer-token`) is populated or not.

ludusrusso:
@prometherion that's nice. So you mean to ignore the `--k8s-api-token ` if the code is in cluster and check for it only when is running out of cluster! Nice :D

Which is the better way to check if the pod is running in cluster?
I can use the 

```go
rest.InClusterConfig()
```

but not sure if this is the best way!

What do you think of naming the flag `--k8s-bearer-token`? Since we are out of cluster I prefer to add the k8s prefix to add context!

davideimola:
I agree with @prometherion `--bearer-token` is way better! Maybe it can be `--override-bearer-token` too...

MaxFedotov:
And what do you say about using a single parameter for both `k8s-control-plane-url` and `--k8s-api-token`, for example a path for a `kubeconfig` file, like it is done in many different operators (like `cluster-autoscaler`):
```go
kubeConfigFile = flag.String("kubeconfig", "", "Path to kubeconfig file with authorization and master location information.")

...

func getKubeConfig() *rest.Config {
	log.V(4).Info("Getting kubeconfig")

	if *kubeConfigFile != "" {
		log.V(1).Info("Building kubeconfig from file", "filename", *kubeConfigFile)
		// use the current context in kubeconfig
		cfg, err := clientcmd.BuildConfigFromFlags("", *kubeConfigFile)
		if err != nil {
			klog.Fatalf("Failed to build config: %v", err)
		}
		return cfg
	}

	cfg, err := rest.InClusterConfig()
	if err != nil {
		klog.Fatalf("Failed to build Kubernetes configuration: %v", err)
	}

	return cfg
}
```

ludusrusso:
@MaxFedotov this is an option that I tested, it is already enabled by `ctrl.GetConfigOrDie()` that, according to docs, read a `--kubeconfig` option for non in cluster stuff.

```
	// GetConfig creates a *rest.Config for talking to a Kubernetes apiserver.
	// If --kubeconfig is set, will use the kubeconfig file at that location.  Otherwise will assume running
	// in cluster and use the cluster provided kubeconfig.
	//
	// Config precedence
	//
	// * --kubeconfig flag pointing at a file
```

However it does not work if you are authentication to the cluster in kubeconfig using certificate, which is the standard.

ludusrusso:
@prometherion I updated the cose checking if it is running in cluster or not!

prometherion:
> However it does not work if you are authentication to the cluster in kubeconfig using certificate, which is the standard.
> 
> To do this we need to enable the proxy to authenticate to the cluster both using bearerToken and certificates. But I do not know how complex is this to implement.

We're tracking this on #67: if it's feasible we can close this (or reuse part of the code) in favor of the new PR.

ludusrusso:
Yes! Closing...

PR #68, opened 2021-03-17
Using kubeconfig to access the kubernetes api

### LINKED ISSUE #67: Run in out-of-cluster mode
This idea started while investigating the code implementing #66 along with @ludusrusso.

Actually, `capsule-proxy` is just designed to run in the __in cluster__ configuration: this means the controller-runtime client will be feed with the CA certificate, the Bearer Token, and other options provided by the ServiceAccount volume mounted in the Pod.

The Bearer Token is definitely an important piece in the whole picture since it's the grant to get all the RBAC capabilities, as the user impersonification:

https://github.com/clastix/capsule-proxy/blob/888cab3ff2e233f330ec787626fc39643ca39f11/internal/webserver/webserver.go#L289-L291

Although this works in the __in cluster__ configuration, it's not the same for the __out of cluster__ mode: the reason is that a `kubeconfig` file is provided that could use the authentication based on a certificate rather than the JWT one.

We should investigate if we can use the Reverse Proxy transport to support the certificate authentication (if provided) and fallback to the bearer token: essentially, what we have to do is to set up the Transport for the Reverse Proxy to use the certificate if provided.

https://github.com/clastix/capsule-proxy/blob/888cab3ff2e233f330ec787626fc39643ca39f11/internal/webserver/webserver.go#L35-L37

With this, we're guaranteeing the ability to run also in the out-of-cluster mode, in production, or debug mode at end-user wish!

ludusrusso:
This PR Closes #67 and allow the connection to kubernetes using kubectl files.

## Changes

1. `k8s-control-plane-url` cmd option is Deprecated, now k8s api endpoint is obtained by kubeconfig, 
2. Authentication to k8s is done using certificates from kubeconfig. 
3. If running out of cluster you should provide the kubeconfig file using `--kubeconfig` cmd option!

## Usage

```bash
go run main.go --ssl-cert-path=/tmp/localhost.pem --ssl-key-path=/tmp/localhost-key.pem  --kubeconfig=./config.kub --enable-ssl=true
```

## Tests

I've manually tested with kind and I'm able to get nemespaces kubernetes api both out of cluster and in cluster.
In cluster it is using the TLS configuration provided by kubernetes automatically!

@prometherion when tests are ready we should run this against automated tests before merging.

@MaxFedotov  @bsctl any feedback?

Thanks to @davideimola for helping!

### DIFF
--- a/internal/options/kube.go
+++ b/internal/options/kube.go
@@ -2,8 +2,6 @@ package options
 
 import (
 	"context"
-	"crypto/tls"
-	"crypto/x509"
 	"fmt"
 	"net"
 	"net/http"
@@ -11,7 +9,7 @@ import (
 	"time"
 
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/util/cert"
+	"k8s.io/client-go/transport"
 )
 
 type kubeOpts struct {
@@ -22,7 +20,11 @@ type kubeOpts struct {
 }
 
 func NewKube(controlPlaneUrl string, groupName string, claimName string, config *rest.Config) (ListenerOpts, error) {
-	u, err := url.Parse(controlPlaneUrl)
+	host := config.Host
+	if controlPlaneUrl != "" {
+		host = controlPlaneUrl
+	}
+	u, err := url.Parse(host)
 	if err != nil {
 		return nil, fmt.Errorf("cannot create Kubernetes Options due to failed URL parsing: %s", err.Error())
 	}
@@ -51,7 +53,15 @@ func (k kubeOpts) PreferredUsernameClaim() string {
 	return k.claimName
 }
 
-func (k kubeOpts) ReverseProxyTransport() *http.Transport {
+func (k kubeOpts) ReverseProxyTransport() (*http.Transport, error) {
+	transportConfig, err := k.config.TransportConfig()
+	if err != nil {
+		return nil, err
+	}
+	tlsConfig, err := transport.TLSConfigFor(transportConfig)
+	if err != nil {
+		return nil, err
+	}
 	return &http.Transport{
 		Proxy: http.ProxyFromEnvironment,
 		DialContext: func(ctx context.Context, network, addr string) (conn net.Conn, e error) {
@@ -61,17 +71,6 @@ func (k kubeOpts) ReverseProxyTransport() *http.Transport {
 			}).DialContext(ctx, network, addr)
 		},
 		TLSHandshakeTimeout: 10 * time.Second,
-		TLSClientConfig: &tls.Config{
-			RootCAs: func() (cp *x509.CertPool) {
-				var err error
-				cp, err = cert.NewPool(k.config.CAFile)
-				if err != nil {
-					cp, _ = cert.NewPoolFromBytes(k.config.CAData)
-				}
-				return
-			}(),
-			NextProtos: k.config.NextProtos,
-			ServerName: k.config.ServerName,
-		},
-	}
+		TLSClientConfig:     tlsConfig,
+	}, nil
 }
--- a/main.go
+++ b/main.go
@@ -34,7 +34,7 @@ func main() {
 	var mgr ctrl.Manager
 
 	listeningPort := flag.Uint("listening-port", 9001, "HTTP port the proxy listens to (default: 9001)")
-	k8sControlPlaneUrl := flag.String("k8s-control-plane-url", "https://kubernetes.default.svc", "Kubernetes control plane URL (default: https://kubernetes.default.svc)")
+	k8sControlPlaneUrl := flag.String("k8s-control-plane-url", "", "Kubernetes control plane URL (DEPRECATED)")
 	capsuleUserGroup := flag.String("capsule-user-group", "capsule.clastix.io", "The Capsule User Group eligible to create Namespace for Tenant resources (default: capsule.clastix.io)")
 	usernameClaimField := flag.String("oidc-username-claim", "preferred_username", "The OIDC field name used to identify the user (default: preferred_username)")
 	bindSsl := flag.Bool("enable-ssl", false, "Enable the bind on HTTPS for secure communication (default: false)")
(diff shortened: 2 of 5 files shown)


ludusrusso:
Thinking about it, we cannot skip completely bearer token! Sorry I'm tired :D

prometherion:
> Bearer token authentication is used but in my option we do not need it anymore, @prometherion what do you think?

The token is mandatory for the in cluster mode since the Service Account token is used to authenticate against the API Server.

davideimola:
As I said to @ludusrusso, I think we can cause a breaking changes while removing the flag `--k8s-control-plane-url` if someone uses custom configs.

What do you think @prometherion and @bsctl ?

bsctl:
> As I said to @ludusrusso, I think we can cause a breaking changes while removing the flag `--k8s-control-plane-url` if someone uses custom configs.

@ludusrusso that option is used in the sidecar scenario and removing it is a breaking change. Please take care, we have currently the proxy in production.

prometherion:
> @ludusrusso that option is used in the sidecar scenario and removing it is a breaking change. Please take care, we have currently the proxy in production.

The feature is planned for 0.1.0 and it will be breaking change.

The issue with the commit is that we don't have to edit the Helm Chart options until the 0.1.0 release hasn't yet published: said so, please, drop the commit, otherwise the CI wll publish a new Helm Chart release that will be broken (we've already faced in Capsule).

ludusrusso:
> > As I said to @ludusrusso, I think we can cause a breaking changes while removing the flag `--k8s-control-plane-url` if someone uses custom configs.
> 
> @ludusrusso that option is used in the sidecar scenario and removing it is a breaking change. Please take care, we have currently the proxy in production.

Yes! I understand this. What is the best for you? Actually the options is no more required since it is automatically loaded by `rest.GetConfigOrDie()` function. I just read the kubernetes url from `config` instead of from the option, and it is loaded correctly both in cluster and out of cluster.

The problem here is that if you call the program with a option that is not configured in `flag` will make the program fails, as in capsule/issues/232

So now if I leave the options in chart the chart will fails. We have two option:
1. Leave the option  both in chart and in code but ignore it
2. Leave the option both in chart and in code and, if present, use it instead of the one coming from config.

What do you prefer?

prometherion:
We'll deal with the chart once releasing 0.1.0, we have to consider the main branch (actually, `master`) as the development one.

Said so, let's ignore the chart update as well removing the CLI flag `--k8s-control-plane-url`.

prometherion:
> Said so, let's ignore the chart update as well removing the CLI flag --k8s-control-plane-url.

Ok, said a stupid thing: let's mark the flag as deprecated, otherwise, the Helm chart would break upon installation during the CI.

We're going to open an issue to remove the deprecated flag on 0.1.0 release: this will require a little more effort on your side @ludusrusso since it means allowing also the deprecated way, hope you can understand.

main.go
```diff
@@ -34,7 +34,6 @@ func main() {
 	var mgr ctrl.Manager
 
 	listeningPort := flag.Uint("listening-port", 9001, "HTTP port the proxy listens to (default: 9001)")
-	k8sControlPlaneUrl := flag.String("k8s-control-plane-url", "https://kubernetes.default.svc", "Kubernetes control plane URL (default: https://kubernetes.default.svc)")
```
prometherion, line 37: We have to keep this for backward compatibility, planning to remove starting from 0.1.0.

Let's default to an empty string and add a log informing the end-user we're going to deprecate this in the upcoming release. If the value is provided we'll use it.

prometherion (changes requested):

internal/options/kube.go
```diff
@@ -2,16 +2,14 @@ package options
 
 import (
 	"context"
-	"crypto/tls"
-	"crypto/x509"
 	"fmt"
+	"k8s.io/client-go/transport"
```
prometherion, line 6: Imports statement is not sorted: should be placed in the underlying block along with `k8s.io/client-go/rest`

prometherion (changes requested):

ludusrusso:
> > Said so, let's ignore the chart update as well removing the CLI flag --k8s-control-plane-url.
> 
> Ok, said a stupid thing: let's mark the flag as deprecated, otherwise, the Helm chart would break upon installation during the CI.
> 
> We're going to open an issue to remove the deprecated flag on 0.1.0 release: this will require a little more effort on your side @ludusrusso since it means allowing also the deprecated way, hope you can understand.

No problem for that. So, If I've got it correctly:

1. I leave the `k8s-control-plane-url` flag in code as "deprecated". 
2. So in case the flag is passed as cmd line option, I'll use it in place of `k.config.Host`.
3. I do not change the Charts now! But
4. I remove the flag from yaml spec!

I'm not sure how to deprecate the flag. I was thinking on adding a log in case the user is using it and change the description in 
 `Kubernetes control plane URL (DEPRECATED)`. Is there a better way?
 
 Thanks for the help!

prometherion:
> 1. I leave the `k8s-control-plane-url` flag in code as "deprecated".
> 2. So in case the flag is passed as cmd line option, I'll use it in place of `k.config.Host`.
> 3. I do not change the Charts now! But
> 4. I remove the flag from yaml spec!

If you're referring to the Charts values, we have to keep it.

> I'm not sure how to deprecate the flag. I was thinking on adding a log in case the user is using it and change the description in
> `Kubernetes control plane URL (DEPRECATED)`. Is there a better way?

It's good enough, thanks!

ludusrusso:
> If you're referring to the Charts values, we have to keep it.
> 
> >

Nono, I mean from pure k8s files `deploy/*yaml`

prometherion:
> I mean from pure k8s files `deploy/*.yaml`

Let's keep them, we'll tackle them along with the Helm release.

internal/options/kube.go
```diff
@@ -21,8 +19,13 @@ type kubeOpts struct {
 	config    *rest.Config
 }
 
+// TODO: Remove k8sControlPlaneUrl
```
prometherion, line 22: `TODO`s are something I really hate in code: we can ignore it since we're already tracking with #71 :)

main.go
```diff
@@ -34,7 +34,8 @@ func main() {
...
+	// TODO: Remove k8s-control-plane-url
```
prometherion, line 37: Same here, already tracking on #71.

prometherion (changes requested):
Final change requests but I think we can merge this!

Just a little remark, although non documented: could you amend commits following the semantic commit message styleguide?

Something as:

```
doc: running capsule locally
refactor!: transport configuration from k8s REST configuration
```

Obviously, commit subjects are just an example, feel free to edit at your wish since you're the author!

main.go
ludusrusso: Ok sorry for that!

ludusrusso:
> Final change requests but I think we can merge this!
> 
> Just a little remark, although non documented: could you amend commits following the semantic commit message styleguide?
> 
> Something as:
> 
> ```
> doc: running capsule locally
> refactor!: transport configuration from k8s REST configuration
> ```
> 
> Obviously, commit subjects are just an example, feel free to edit at your wish since you're the author!

Now should be ok!

prometherion (approved):
LGTM! Thanks for your effort @ludusrusso! 🚀

PR #69, opened 2021-03-17
e2e on Kubernetes matrix with GitHub Actions

### LINKED ISSUE #64: Setting up e2e tests
Would be great having an e2e test suite to check all the functionalities of `capsule-proxy`.

With `capsule`, we're relying on Ginkgo but I think it would be overkill: we don't have CRDs, we just need to check the reverse proxy is working as expected, what we just need is a simple Kubernetes cluster pointing to.

Besides the retrieval of expected and filtered items, we have to check three functions:
1. the JSON output
2. the Table output `kubectl`
3. each of the above with the `--watch/-w` flag

Implementing all of them could be challenging, especially the `watch` one since there are WebSockets involved, but I guess we can hack bash with some timeouts to check the output: this can be postponed if considered too much tricky, the first purpose of this issue is to avoid merging bugs and enforce the reliability of the main branch.

Since it's just a matter of expectation testing, [`BATS`](https://github.com/sstephenson/bats) could be really interesting and easy to use.

The e2e test suite should be run against several Kubernetes versions, as well as [the Capsule supported one](https://github.com/clastix/capsule/blob/56adfe6a35d2271f1556d794614007aee656dbb5/.github/workflows/e2e.yml#L14), installing it using the [official Helm repository](https://github.com/clastix/capsule/tree/master/charts/capsule) along with the `capsule-proxy` one.

prometherion:
Closes partially #64, this is just a scaffolding but good enough to be safe with further enhancements.

Can be still improved.


prometherion:
We got e2e testing! 🎉 

image

PR #79, opened 2021-03-22
Switching over middleware to solve out-of-cluster mode

### LINKED ISSUE #65: Switch over HTTP middlewares
We got several middlewares in the current codebase and strictly related to #54 once closed:

- [`isNamespaceListing`](https://github.com/clastix/capsule-proxy/blob/c3d7fda8da2aa7e006c0593aa4469355643b4a57/internal/webserver/webserver.go#L214-L221) checking if the request is a Namespace listing request, this can be safely dropped
- [`isNodeListing`](https://github.com/clastix/capsule-proxy/blob/c3d7fda8da2aa7e006c0593aa4469355643b4a57/internal/webserver/webserver.go#L223-L230) same as above, besides the fact addressing the node listing
- [`checkJwt`](https://github.com/clastix/capsule-proxy/blob/c3d7fda8da2aa7e006c0593aa4469355643b4a57/internal/webserver/webserver.go#L94-L122) that is checking the authentication provided by the bearer token

This last one must be converted to a middleware for all the routes transformed by `capsule-proxy`.

Although not properly a middleware, also the [`reverseProxyFunc`](https://github.com/clastix/capsule-proxy/blob/c3d7fda8da2aa7e006c0593aa4469355643b4a57/internal/webserver/webserver.go#L280-L297) could be considered a middleware, although performed after the single resource manipulation (`namespaces`, `nodes`, etc.): this could be splitted up in two middleware, the first one adding the __impersonate__ headers, and the second one performing the reverse proxy call.

Since the planned new routes are coming, we must group resources and apply middlewares.

### LINKED ISSUE #74: The proxy does not work when running with TSL authentication in out of cluster mode
When running on out of cluster mode, `bearerToken` is empty. In this case, we have to skip impersonification here

https://github.com/clastix/capsule-proxy/blob/4a7e4c60479f8c1b372d8e43bd28b6183c3e2a45/internal/webserver/webserver.go#L298-L303

@prometherion I do not understand exactly why this is working in cluster mode but not in out of cluster mode. In my understanding, when the proxy performs namespace filtering, it should skip completely end user impersonification.

prometherion:
Closes #74, and closes #65 as well.

Along with the additional test case checking for the expected failure due to missing RoleBinding for the _default_ namespace, this refactoring solves the out-of-cluster mode, as well implementing the middleware logic that skips User-Authentication for the decorated requests (`/api/v1/namespaces` and `/api/v1/nodes`).

A prerequisite to solving #74 was implementing #65, although already assigned to @ludusrusso: having a review from you would be great, as well co-authoring the commits.

### DIFF
--- a/internal/webserver/webserver.go
+++ b/internal/webserver/webserver.go
@@ -97,8 +97,8 @@ func (n *kubeFilter) InjectClient(client client.Client) error {
 	return nil
 }
 
-func (n kubeFilter) checkJwt(fn http.HandlerFunc) http.HandlerFunc {
-	return func(writer http.ResponseWriter, request *http.Request) {
+func (n kubeFilter) checkJWTMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
 		var err error
 
 		token := strings.Replace(request.Header.Get("Authorization"), "Bearer ", "", -1)
@@ -123,8 +123,17 @@ func (n kubeFilter) checkJwt(fn http.HandlerFunc) http.HandlerFunc {
 				return
 			}
 		}
-		fn(writer, request)
-	}
+		next.ServeHTTP(writer, request)
+	})
+}
+
+func (n kubeFilter) reverseProxyMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+		next.ServeHTTP(writer, request)
+
+		log.V(5).Info("debugging request", "uri", request.RequestURI, "method", request.Method)
+		n.reverseProxy.ServeHTTP(writer, request)
+	})
 }
 
 func (n kubeFilter) handleRequest(request *http.Request, username string, selector labels.Selector) {
@@ -145,8 +154,10 @@ func (n kubeFilter) handleRequest(request *http.Request, username string, select
 	log.V(4).Info("updating RawQuery", "query", q.Encode())
 	request.URL.RawQuery = q.Encode()
 
-	log.V(4).Info("Updating the token", "token", n.bearerToken)
-	request.Header.Set("Authorization", fmt.Sprintf("Bearer %s", n.bearerToken))
+	if len(n.bearerToken) > 0 {
+		log.V(4).Info("Updating the token", "token", n.bearerToken)
+		request.Header.Set("Authorization", fmt.Sprintf("Bearer %s", n.bearerToken))
+	}
 }
 
 func (n kubeFilter) nodesHandler(writer http.ResponseWriter, request *http.Request) {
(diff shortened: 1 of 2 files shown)


internal/webserver/webserver.go
```diff
@@ -145,8 +154,10 @@ func (n kubeFilter) handleRequest(request *http.Request, username string, select
 	log.V(4).Info("updating RawQuery", "query", q.Encode())
 	request.URL.RawQuery = q.Encode()
 
-	log.V(4).Info("Updating the token", "token", n.bearerToken)
-	request.Header.Set("Authorization", fmt.Sprintf("Bearer %s", n.bearerToken))
+	if len(n.bearerToken) > 0 {
```
prometherion, line 157: To avoid interferences with the empty string due to run on out-of-cluster mode, we're skipping the update of the token.

Running in out-of-cluster mode, the transport will inherit the cluster administrator certificates, so we're good from the authorization perspective.

internal/webserver/webserver.go
```diff
@@ -217,40 +228,45 @@ func (n kubeFilter) namespacesHandler(writer http.ResponseWriter, request *http.
...
 	r := mux.NewRouter()
-	r.HandleFunc("/api/v1/namespaces", func(writer http.ResponseWriter, request *http.Request) {
-		n.isNamespaceListing(n.checkJwt(n.namespacesHandler))(writer, request)
-		n.reverseProxyFunc(writer, request)
-	})
-	r.HandleFunc("/api/v1/nodes", func(writer http.ResponseWriter, request *http.Request) {
-		n.isNodeListing(n.checkJwt(n.nodesHandler))(writer, request)
-		n.reverseProxyFunc(writer, request)
-	})
+	r.StrictSlash(true)
```
prometherion, line 233: We're basically avoiding dealing with special cases with the trailing slash of paths.

internal/webserver/webserver.go
```diff
@@ -217,40 +228,45 @@ func (n kubeFilter) namespacesHandler(writer http.ResponseWriter, request *http.
...
+	r.Use(n.reverseProxyMiddleware)
```
prometherion, line 234: I love this middleware, because in the end, we're telling each route to perform the reverse-proxy `ServeHTTP` method, delegating to the handlers just the request manipulation: less code is better.

internal/webserver/webserver.go
```diff
@@ -217,40 +228,45 @@ func (n kubeFilter) namespacesHandler(writer http.ResponseWriter, request *http.
...
+
+	node := r.Path("/api/v1/nodes").Subrouter()
+	node.Use(n.checkJWTMiddleware)
+	node.HandleFunc("", func(writer http.ResponseWriter, request *http.Request) {
+		n.nodesHandler(writer, request)
+	})
+
 	r.PathPrefix("/").HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
-		n.reverseProxyFunc(writer, request)
+		if n.serverOptions.IsListeningTls() {
```
prometherion, line 254: Just for the catch-all route (so the ones non overloaded by our reverse-proxy) we're doing the _impersonification_: this is required since the reverse proxy inherited the transport from the cluster administrator `kubeconfig` (or the equivalent) and the following requests should be presented on behalf of the requester, and _impersonification_ is exactly doing this.

ludusrusso:
That's super nice! I suggest just two small improvements..

internal/webserver/webserver.go
```diff
@@ -217,40 +228,45 @@ func (n kubeFilter) namespacesHandler(writer http.ResponseWriter, request *http.
...
+	ns.Use(n.checkJWTMiddleware)
+	ns.HandleFunc("", func(writer http.ResponseWriter, request *http.Request) {
+		n.namespacesHandler(writer, request)
+	})
+
+	node := r.Path("/api/v1/nodes").Subrouter()
+	node.Use(n.checkJWTMiddleware)
+	node.HandleFunc("", func(writer http.ResponseWriter, request *http.Request) {
+		n.nodesHandler(writer, request)
+	})
```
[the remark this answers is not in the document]
prometherion: You're anticipating a feature I'd like to introduce in the upcoming commits, it's the module concept.

I was thinking of an interface to register a said module (as `NodeList`, `NodeGet`, `NamespaceList`, and more to come) which encapsulates the underlying business logic. There would be a little of copy-paste between modules but it's a trade-off.

The Capsule Proxy will just need to register the modules with the list of required middleware (right now, just the JWT check) with this pseudo-code:

```go
mr := r.PathPrefix("/").SubRouter()
mr.Use(n.checkJWTMiddleware)

for _, m := range modules {
  m.Register(mr)
}
```

Just wanted to keep the PR as simple as possible for the review but any improvement, both on code or feature design, is highly welcomed.

internal/webserver/webserver.go
```diff
@@ -217,40 +228,45 @@ func (n kubeFilter) namespacesHandler(writer http.ResponseWriter, request *http.
...
+				r := req.NewHttp(request, n.usernameClaimField)
+				username, groups, err := r.GetUserAndGroups()
+				if err == nil {
+					log.V(4).Info("Impersonating for the current request", "username", username, "groups", groups, "token", n.bearerToken)
+					if len(n.bearerToken) > 0 {
+						request.Header.Set("Authorization", fmt.Sprintf("Bearer %s", n.bearerToken))
+					}
+					request.Header.Add("Impersonate-User", username)
+					request.Header.Add("Impersonate-Group", strings.Join(groups, ","))
+				}
```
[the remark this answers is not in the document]
prometherion: Implemented with 925224e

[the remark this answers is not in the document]
ludusrusso: Way better no!

PR #96, opened 2021-04-07
chore: node listing should rely on tenant's nodeSelector spec

### LINKED ISSUE #52: Node listing should rely on the Tenant's nodeSelector spec
Starting from capsule-proxy v0.0.5, it is possible to enable node listing for tenant owners through the `capsule.clastix.io/enable-node-listing` Tenant annotation.

The actual implementation requires a node to be explicitly labelled as `capsule.clastix.io/tenant=<Tenant Name>` to be listed by the corresponding Tenant owners. This implementation choice comes with some flaws:
- It does not support a scenario in which the same node is associated with multiple tenants
- It requires the cluster admin to explicitly add the `capsule.clastix.io/tenant=<Tenant Name>` to each node in the cluster, but this step is not documented
- It is totally unaware of the Tenant's `nodeSelector` spec. If a node is labelled as `capsule.clastix.io/tenant=<Tenant Name>` , tenant owners are able to list that node even if it violates the `nodeSelector` policy, i.e. even if they are not able to schedule pods on that node. Viceversa, if a node satisfies the `nodeSelector` policy but is not labelled as `capsule.clastix.io/tenant=<Tenant Name>`, users can schedule pods on it but are not able to list it.

It would be better to avoid configuration fragmentation for the node visibility feature: the same rule (preferably the `nodeSelector` spec, which allows node sharing among multiple tenants) should control both the node visibility in the capsule-proxy and the possibility to schedule pods on it.

MaxFedotov:
closes #52

### DIFF
--- a/internal/webserver/nodes.go
+++ b/internal/webserver/nodes.go
@@ -6,6 +6,9 @@ import (
 	"net/http"
 	"strconv"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/selection"
+
 	capsulev1alpha1 "github.com/clastix/capsule/api/v1alpha1"
 	"github.com/gorilla/mux"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


internal/webserver/nodes.go
```diff
@@ -32,82 +35,112 @@ func (n kubeFilter) registerNode(router *mux.Router) {
...
+					matches++
+				}
+			}
+			if matches == len(selector) {
+				names = append(names, node.GetName())
+			}
+		}
+	}
+	if len(names) > 0 {
+		return labels.NewRequirement("kubernetes.io/hostname", selection.In, names)
```
prometherion, line 54: I guess we're doing this due to listing, right?

My idea was to construct essentially the LabelSelector adding barely the `map[string]string` as _Requirement_ but that wouldn't work.

If it is, LGTM.

MaxFedotov: yes, because a user can own multiple tenants and each of them may have a different `nodeSelector`.

prometherion: Great, going to merge this, although squashing the commit to change the message.

Issue #99, opened 2021-04-10
Add support for k8s webhook token authentication strategy


gautam2187:
Currently capsule-proxy assumes that the bearer token is a valid JWT token. This may not be the case with k8s clusters using webhook token authentication strategy as mentioned here https://kubernetes.io/docs/reference/access-authn-authz/authentication/#webhook-token-authentication

In this case bearer token could be a non JWT custom token that the webhook can process.

Currently webserver.go#checkJWTMiddleware passes the token to TokenReview and gets a populated TokenReview but the populated groups are not used down the chain. So, it fails the next check in webserver.go#checkUserInCapsuleGroupMiddleware.

prometherion:
Thanks for reporting this @gautam2187, a nice enhancement since we weren't taking care of this strategy: going to add this to the 0.1.0 milestone!

PR #100, opened 2021-04-11
feat: support for webhook token authentication strategy

gautam2187:
#99

### DIFF
--- a/internal/request/http.go
+++ b/internal/request/http.go
@@ -1,11 +1,14 @@
 package request
 
 import (
+	"context"
 	"fmt"
 	h "net/http"
 	"strings"
 
 	"github.com/dgrijalva/jwt-go"
+	authenticationv1 "k8s.io/api/authentication/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type authType int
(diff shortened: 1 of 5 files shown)


internal/request/http.go
```diff
@@ -58,7 +77,14 @@ func (h http) bearerToken() string {
 	return strings.ReplaceAll(h.Header.Get("Authorization"), "Bearer ", "")
 }
 
+func (h http) isNonJwtBearerToken() bool {
+	return len(h.bearerToken()) > 0 && !h.isJwtToken()
+}
+
 func (h http) getAuthType() authType {
+	if h.isNonJwtBearerToken() {
```
prometherion, line 85: Honestly, since cases are increasing, I would go for something like this:

```go
switch {
case isJwtToken():
  return bearerBased
case len(h.bearerToken()) > 0:
  return bearerBased
case h.TLS != nil:
  return certificateBased
default:
  return anonymousBased
}
```

With this, if it's not a JWT but the bearer token is there, definitely it's a bearer-based authentication so we can avoid putting an additional function (`isNonJwtBearerToken`), WDYT?

prometherion (changes requested):
Hey, thanks for this PR, really appreciated it since we were totally missing this authentication provider!

I suggested a few changed, can you share your thoughts? If it's ok, we can get it merged for the upcoming 0.1.0 releas

prometherion:
Oh, forgot to ask if you rename the commit message according to [conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/).

I would suggest something as: `feat: support for webhook token authentication strategy`

internal/request/http.go
```diff
@@ -1,28 +1,33 @@
 package request
 
 import (
+	"context"
 	"fmt"
 	h "net/http"
 	"strings"
 
 	"github.com/dgrijalva/jwt-go"
+	v1 "k8s.io/api/authentication/v1"
```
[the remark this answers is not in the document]
gautam2187: done!

internal/request/http.go
gautam2187: makes sense, simplified it further, LMK if this looks better.

prometherion (approved):
Thanks for the changes, definitely LGTM!

PR #105, opened 2021-04-15
refactor!: remove --k8s-control-plane-url CLI flag from Helm Chart

### LINKED ISSUE #71: Removal of --k8s-control-plane-url CLI flag
With #68 we're totally relying on the configuration provided by the Service Account options (when running in in-cluster mode) or the provided `kubeconfig` file (when running in out-of-the-cluster mode).

The code actually warns users running the development branch of the deprecation policy: for 0.1.0 this has to be removed since there's any more need for it.

# Tasks

- [ ] [Update Helm chart](#105)
- [x] [Update code](#107)

ludusrusso:
Related to #71

### DIFF
--- a/charts/capsule-proxy/Chart.yaml
+++ b/charts/capsule-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: 0.0.5
+appVersion: 0.1.0
 description: A Helm chart for Kubernetes
 name: capsule-proxy
 type: application
--- a/charts/capsule-proxy/Chart.yaml
+++ b/charts/capsule-proxy/Chart.yaml
@@ -3,4 +3,4 @@ appVersion: 0.0.5
 description: A Helm chart for Kubernetes
 name: capsule-proxy
 type: application
-version: 0.1.3
+version: 0.1.4
--- a/charts/capsule-proxy/templates/daemonset.yaml
+++ b/charts/capsule-proxy/templates/daemonset.yaml
@@ -47,7 +47,6 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         args:
         - --listening-port={{ .Values.options.listeningPort }}
-        - --k8s-control-plane-url={{ .Values.options.k8sControlPlaneUrl }}
         - --capsule-user-group={{ .Values.options.capsuleUserGroup }}
         - --zap-log-level={{ .Values.options.logLevel }}
         - --enable-ssl={{ .Values.options.enableSSL }}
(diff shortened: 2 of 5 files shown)


prometherion (changes requested):
We're missing the Helm Chart version bump.

ludusrusso:
I don't remember if I need to remove the flag from the code or only from the helm chart

internal/options/kube.go
```diff
@@ -20,11 +20,8 @@ type kubeOpts struct {
 	config    *rest.Config
 }
 
-func NewKube(controlPlaneURL string, groupName string, claimName string, config *rest.Config) (ListenerOpts, error) {
+func NewKube(groupName string, claimName string, config *rest.Config) (ListenerOpts, error) {
```
prometherion, line 23: I don't want to be pedantic but can I ask you to split changes into several commits?

One for the Helm charts, another for the Go code-base changes.

ludusrusso: Yep, but I'm not sure if we should do this!

prometherion:
> I don't remember if I need to remove the flag from the code or only from the helm chart

Well, you raised a good topic.

I think we should create two MRs: one for the code, the second one for the Charts.

Saying this since we have to release, first, the new binary along with Docker image, otherwise, we could end up with errors for users installing in the release process.

ludusrusso:
> > I don't remember if I need to remove the flag from the code or only from the helm chart
> 
> Well, you raised a good topic.
> 
> I think we should create two MRs: one for the code, the second one for the Charts.
> 
> Saying this since we have to release, first, the new binary along with Docker image, otherwise, we could end up with errors for users installing in the release process.

Got it so this one is only for Helm chart!

charts/capsule-proxy/Chart.yaml
```diff
@@ -3,4 +3,4 @@ appVersion: 0.0.5
 description: A Helm chart for Kubernetes
 name: capsule-proxy
 type: application
-version: 0.1.3
+version: 0.1.4
```
ludusrusso, line 6: @prometherion is this right?

prometherion: LGTM

prometherion:
@ludusrusso may I ask you to refer the `appVersion` to v0.1.0?

Once done, we can merge this!

ludusrusso:
> @ludusrusso may I ask you to refer the `appVersion` to v0.1.0?
> 
> Once done, we can merge this!

done!

prometherion (approved):
LGTM!

PR #107, opened 2021-04-15
refactor!: remove k8s-control-plane-url CLI flag

ludusrusso:
Related to #71

### DIFF
--- a/internal/options/kube.go
+++ b/internal/options/kube.go
@@ -20,13 +20,8 @@ type kubeOpts struct {
 	config    *rest.Config
 }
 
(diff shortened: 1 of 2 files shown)


ludusrusso:
@prometherion to pass this we need first to merge #105, the problem here is that tests does not pass due to the flag present in helm chart conf

prometherion:
> @prometherion to pass this we need first to merge #105, the problem here is that tests does not pass due to the flag present in helm chart conf

We can keep these failed, it's not a problem since with #105 we're going to solve this.

PR #109, opened 2021-04-21
docs: typo on README.md

### LINKED ISSUE #108: Documentation need checks
There are some minor errors in the main documentation that need attention

ptx96:
Fixed some typos about Tenant annotations and a little error on the chart doc
Related to #108

### DIFF
--- a/internal/options/kube.go
+++ b/internal/options/kube.go
@@ -20,13 +20,8 @@ type kubeOpts struct {
 	config    *rest.Config
 }
 
-func NewKube(controlPlaneURL string, groupName string, claimName string, config *rest.Config) (ListenerOpts, error) {
-	host := config.Host
-	if controlPlaneURL != "" {
-		host = controlPlaneURL
-	}
-
-	u, err := url.Parse(host)
+func NewKube(groupName string, claimName string, config *rest.Config) (ListenerOpts, error) {
+	u, err := url.Parse(config.Host)
 	if err != nil {
 		return nil, fmt.Errorf("cannot create Kubernetes Options due to failed URL parsing: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


charts/capsule-proxy/README.md
```diff
@@ -54,7 +54,7 @@ The values in your overrides file `myvalues.yaml` will override their counterpar
 
 If you only need to make minor customizations, you can specify them on the command line by using the `--set` option. For example:
 
-        $ helm install capsule-proxy clastix/capsule-proxy --values myvalues.yaml -n capsule-system
+        $ helm install capsule-proxy clastix/capsule-proxy --set "kind=DaemonSet" -n capsule-system
```
prometherion, line 57: Why defining the `kind` value?

Default is `Deployment` but I'm missing the point on why we should document to set the deployment kind to `DaemonSet`.

prometherion (approved):
Just a minor question, but overall LGTM.

prometherion:
Let's ignore the failed checks, we're going to fix this in a while once v0.1.0 is release.

charts/capsule-proxy/README.md
ptx96: Just to provide an example about `--set` option ;
feel free to swap it with another helm chart property (listening port?).

prometherion: We don't have to provide this kind of example since we're considering operators know how to use Helm.

Please, amend the commit dropping this addition.

prometherion: Ah, my bad: it's good enough! :)

prometherion (commented):
Ok, definitely LGTM, just a little remark.

@ptx96 could you split the commit into two parts? One for the Helm Char README, and another one with the core fixes.

PR #116, opened 2021-05-07
switch from command to args

### LINKED ISSUE #115: switching from command to args in deployment examples
switching from `command` to `args` in in deployment examples

bsctl:
close #115


prometherion (approved):
LGTM, just one remark: can you amend commit message using the Semantic Commit Message convention? It would be `chore: switch from command to args...`

PR #117, opened 2021-05-13
Use multiple Impersonate-Group headers

ThatsMrTalbot:
The kubernetes documentation specifies group information is set in multiple headers, not a single header with comma separation. https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation

This change replaces the string.Join with a simple loop over the groups to implement this change.


prometherion (changes requested):
I saw `golangci-lint` is failing, could you fix it according to the requested coding style guides?

But overall, LGTM!

prometherion (approved):

PR #132, opened 2021-08-16
feat: bumping to new capsule version

### LINKED ISSUE #130: Updating to latest changes on v1beta1
As mentioned by @MaxFedotov on clastix/capsule#366, we have to update capsule-proxy to use the latest `v1beta1` version in order to avoid unmarshaling errors.

alegrey91:
Close #130


prometherion:
@alegrey91 reason for failing CI was due to a similar bug we had on Capsule with the Kind action which was missing some dependencies on the virtual machine image we're using `ubuntu:latest`.

Once fixed, the test cases were not aligned to the changes introduced with the latest API changes that occurred during the last commits, I'm referring to the `/spec/namespaceOptions` and `/spec/ingressOptions`.

And finally, updated also the `README.md` with the examples up to last `v1beta1` changes, adding also the documentation for the `PriorityClass` resources I missed implementing.

PR #140, opened 2021-09-03
Ignoring certain user groups from filtering

### LINKED ISSUE #119: Skip the filter when requests come from users with cluster admin permissions
It would be nice to cover the case when the tenant owner is also a cluster admin. Let say the tenant owner belongs to two groups: `devops` and `admin`.  A tenant has been assigned to the `devops` group. Also the `admin` group has permission of cluster admin.

Right now, when the request coming from such tenant owner is intercepted by the proxy, because of the `--capsule-user-group` and then the filter is applied preventing the request to go straight to APIs server.

prometherion:
Closes #119.

Adding a new slice CLI flag named `--ignored-user-group` an administrator can set which groups must be ignored from filtering, executing just the impersonate handler and bypassing all the underlying handlers.

### DIFF
--- a/charts/capsule-proxy/Chart.yaml
+++ b/charts/capsule-proxy/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 appVersion: 0.1.0
-description: A Helm chart for Kubernetes
+description: Helm Chart for Capsule Proxy, addon for Capsule, the multi-tenant Operator
 name: capsule-proxy
 type: application
 version: 0.1.4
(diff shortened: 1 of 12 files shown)


bsctl (commented):
missing `--ignored-user-group` option in daemonset template of the chart

```
        {{- range .Values.options.ignoredUserGroups }}
        - --ignored-user-group={{ . }}
        {{- end}}
```

bsctl (approved):
lgtm

PR #142, opened 2021-09-17
docs: including metrics url

mendrugory:
Including metrics url and also sorting them


prometherion (approved):
I missed this on the previous PR, well done!

PR #151, opened 2021-09-24
regular users should be able to get namespace list

### LINKED ISSUE #149: Users for whom tenant owners created rolebindings in their tenant namespaces should be able to get them using kubectl
In a current `capusle-proxy` version, if `Alice`, an owner of a tenant `oil` had created rolebindings for `Joe` (an ordinary kubernetes user) inside her tenants namespaces, `Joe` won't be able to see them using `kubectl get ns`.

That is a problem, especially in big organizations, where tenants on production are usually owned by devops or sre teams and they had to create restricted roles in some of their namespaces for members of dev\qa teams.

After discussion with @prometherion, we came with an idea to modify the current way of showing namespaces in `capsule-proxy`. Instead of relying on getting tenant information from request `UserInfo` and then adding `capsule.clastix.io/tenant` label to request, we can create a separate rolebinding controller (the same way as it is done in `CapsuleConfiguration` controller), which will store mapping of users and the namespaces, where they had rolebindings in some structure, which will be used to filter out namespaces when user performs `kubectl get ns`

MaxFedotov:
closes #149

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,8 @@ e2e/%: docker-build
 	helm repo add clastix https://clastix.github.io/charts
 	helm upgrade --install --create-namespace --namespace capsule-system capsule clastix/capsule \
 		--set "manager.resources=null" \
-		--set "manager.options.forceTenantPrefix=true"
+		--set "manager.options.forceTenantPrefix=true" \
+		--set "options.logLevel=8"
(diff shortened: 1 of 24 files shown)


prometherion (approved):
Thanks for this awesome PR, definitely going to provide more reliable and efficient proxying!

PR #158, opened 2021-10-06
Blocking unauthenticated requests when using plain HTTP mode

### LINKED ISSUE #156: Potential security loophole when proxying with wrong config
This is not a vulnerability report even the title sounds like that.

Let me walk it through the process and you'll know why I say that.

## 1. Launch the `capsule-proxy`:

Walk through the steps by following the doc to prepare the below stuff:
- Deploy Capsule
- Generate TLS cert/key by `mkcert`

Then launch the `capsule-proxy` outside of the cluster (it would be the same if you launch it in cluster):
```sh
$ go run . \
  --ssl-cert-path=localhost.pem \
  --ssl-key-path=localhost-key.pem \
  --zap-log-level=10 \
  --enable-ssl=false
```

Please note here that we intentionally disable the TLS with `--enable-ssl=false` as this is the key point.

## 2. Update `alice`'s kubeconfig file

Let assume you've created a Tenant named `gas` and generated `alice`'s kubeconfig by following `Capsule`'s doc.

Now we make a copy of it and rename it as ``, with below update:

```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: <OMITTED>
    server: http://localhost:9001
  name: kind-kind-capsule
contexts:
- context:
    cluster: kind-kind-capsule
    user: alice
  name: alice-gas
current-context: alice-gas
kind: Config
preferences: {}
users:
- name: alice
  user:
    client-certificate: alice-gas.crt
    client-key: alice-gas.key
```

Note the `server: http://localhost:9001` part.

## 3. Have a try:

```sh
# Without proxy, alice can't get namespaces, as expected in K8s
$ KUBECONFIG=alice-gas.kubeconfig kg ns
Error from server (Forbidden): namespaces is forbidden: User "alice" cannot list resource "namespaces" in API group "" at the cluster scope

# With the proxy, alice CAN list all namespaces which is far away from what she should be granted
$ KUBECONFIG=alice-gas-proxy.kubeconfig kg ns
NAME                 STATUS   AGE
capsule-system       Active   25h
default              Active   25h
gas-dev              Active   25h
kube-node-lease      Active   25h
kube-public          Active   25h
kube-system          Active   25h
local-path-storage   Active   25h
```

Note that she should be granted ONLY `gas-dev` namespace in a proxied `kubectl get ns`.

prometherion:
Closes #156, blocking from the beginning proxy start when pointing to a certificate/key pair but in plain HTTP.

A review from @brightzheng100 would be great since he's the reported, thanks for your commitment!


MaxFedotov (approved):
Lgtm!

internal/webserver/middleware/jwt.go
```diff
@@ -41,7 +43,10 @@ func CheckJWTMiddleware(client client.Client, log logr.Logger) mux.MiddlewareFun
 				if statusErr := tr.Status.Error; len(statusErr) > 0 {
 					errors.HandleError(writer, err, "cannot verify the token due to error")
 				}
+			case !tls && len(token) == 0:
```
MaxFedotov, line 46: May be use 'default' here?

brightzheng100:
Sounds a bit weird to me.

1. The TLS is just a mechanism to expose the proxy endpoint for the end users like `alice`;
2. Even TLS is disabled and plain HTTP is used, `alice` is still to provide her certs for K8s authentication, as usual.

But now, when we disable the proxy TLS, `alice`'s cert/key based authentication would not work any more, why?

```sh
$ KUBECONFIG=alice-gas-proxy.kubeconfig kg ns
Error from server (Forbidden): cannot determinate the current user: missing token
```

prometherion:
> 1. The TLS is just a mechanism to expose the proxy endpoint for the end users like `alice`;

Not entirely true: Kubernetes provides several authentication mechanisms. The most popular one is based on token (mostly JWT), a deprecated one (although yet valid) based on certificates. Said so, TLS is used to encrypt communication between client and server, but also for authentication.

> 2. Even TLS is disabled and plain HTTP is used, alice is still to provide her certs for K8s authentication, as usual

This is incorrect: certificate-based authentication is just supported for HTTPS connections.

tl;dr; if you need certificate-based authentication, TLS must be enabled, otherwise you have to deal with JWT, therefore bearer tokens.

brightzheng100:
> TLS is used to encrypt communication between client and server, but also for authentication.

> This is incorrect: certificate-based authentication is just supported for HTTPS connections.

Okay, thanks for the clarifications. @prometherion 

If cert-based authentication must be through TLS, enhancing the error message might be helpful to avoid the potential confusion like what I had (or just me I guess?). 

For example:

```go
errors.HandleUnauthorized(writer, fmt.Errorf("missing token"), "cannot determinate the current user: no cert-based authentication is available when TLS is disabled, and no JWT token is detected.")
```

Well, you can make the message better than what I can think of:)

brightzheng100:
While I'm checking through the relevant lines of code:

```go
func CheckJWTMiddleware(client client.Client, log logr.Logger, tls bool) mux.MiddlewareFunc {
...
				if statusErr := tr.Status.Error; len(statusErr) > 0 {
					errors.HandleError(writer, err, "cannot verify the token due to error")
				}
...
```

1. Should this `err` be `statusErr`, wrapping up with `errors.New(tr.Status.Error)`? -- not sure whether they're the same:

```go
errors.HandleError(writer, statusErr, "cannot verify the token due to error")
```

2. Or should we use this logic instead?

```go
func CheckJWTMiddleware(client client.Client, log logr.Logger, tls bool) mux.MiddlewareFunc {
...
				if tr.Status.Authenticated {
					// and do we need to check further for the user name too?
					//tr.Status.User.Username must be the same as the request claims 
				} else {
					errors.HandleError(writer, errs.New(tr.Status.Error), "cannot authenticate the token due to error")
				}
...
```

Just my 2-cent while learning the code.

prometherion:
> 1. Should this `err` be `statusErr`, wrapping up with `errors.New(tr.Status.Error)`? -- not sure whether they're the same:

Definitely, that would be `nil`, thanks for the hint!

> 2. Or should we use this logic instead?

No need to perform additional check since Kubernetes is already doing any cross-check.

PR #169, opened 2021-10-18
Allowing http with bearer token authentication

### LINKED ISSUE #168: capsule-proxy should support working with ingress controller with tls-termination
It is a quite common scheme for many enterprises to use TLS-termination on an ingress side because it simplifies certificates management.

Mostly these setups use `capsule-proxy` with bearer token authentication, so after #162 was merged it is now impossible to support them and that becomes a backward-incompatible change.

We need to figure a way, how we can support it without compromising overall `capsule-proxy` security.

mendrugory:
it closes #168


mendrugory:
Thanks @MaxFedotov

PR #178, opened 2021-12-29
build(helm): ingress support for kubernetes >= 1.22

### LINKED ISSUE #172: Refine apiVersion's compatibility inside Ingress resource
#  Description

In the capsule-proxy helm charts we manage the `apiVersion` compatibility of the ingress resource based on the target cluster version :

https://github.com/clastix/capsule-proxy/blob/4315b6863d4c1f5b561c6430c4a1f6c6a8569798/charts/capsule-proxy/templates/ingress.yaml#L3-L7

We dropped support for version v1.14.0, so it is no longer needed.

---

At the same time, for cluster version >= v1.14.0, it uses `apiVersion: networking.k8s.io/v1beta1`, which are already unavailable in v1.22+ :

```
Warning: networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
```
so we should adapt the helm charts to the `apiVersion: networking.k8s.io/v1`

1. Spin up a kind cluster ;
2. Install capsule helm charts ;
3. Install capsule-proxy helm charts with Ingress enabled ;
4. Capsule-proxy Ingress works as expected

# Expected behavior

With version \>= v1.12 && < v1.22 
Ingress resource uses `apiVersion: networking.k8s.io/v1`

With version \< v1.12
Ingress resource uses `apiVersion: networking.k8s.io/v1beta1`

prometherion:
Closes #172.

### DIFF
--- a/charts/capsule-proxy/templates/ingress.yaml
+++ b/charts/capsule-proxy/templates/ingress.yaml
@@ -1,9 +1,9 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "capsule-proxy.fullname" . -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
-apiVersion: networking.k8s.io/v1beta1
+{{- if semverCompare ">=1.12-0" .Capabilities.KubeVersion.GitVersion -}}
+apiVersion: networking.k8s.io/v1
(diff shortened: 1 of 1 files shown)


ptx96 (approved):
LGTM

PR #183, opened 2022-01-17
fix: ingress service reference in networking.k8s.io/v1

iwilltry42:
Failed to deploy on k8s v1.22 before this change.
As per https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#httpingresspath-v1-networking-k8s-io

Capsule is great, thank you! :)

### DIFF
--- a/charts/capsule-proxy/templates/ingress.yaml
+++ b/charts/capsule-proxy/templates/ingress.yaml
@@ -1,6 +1,6 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "capsule-proxy.fullname" . -}}
-{{- if semverCompare ">=1.12-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" -}}
 apiVersion: networking.k8s.io/v1
 {{- else -}}
 apiVersion: networking.k8s.io/v1beta1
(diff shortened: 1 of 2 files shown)


charts/capsule-proxy/templates/ingress.yaml
```diff
@@ -31,10 +31,20 @@ spec:
     http:
       paths:
       {{- range .paths }}
+      {{- if semverCompare ">=1.12-0" $.Capabilities.KubeVersion.GitVersion }}
```
prometherion, line 34: Actually, were expecting Capsule to be running on Kubernetes >= 1.16 due to certain constraints, mostly regarding API versions and some admission controllers.

Although the proposed change is optimal, what do you think about putting the said version as check?

prometherion (changes requested):

prometherion:
Thanks, @iwilltry42 for the PR, really appreciated it!

Please, don't forget to bump the Chart version so we can release a new version right away after merging it!

charts/capsule-proxy/templates/ingress.yaml
iwilltry42: I actually changed it to checking for the availability of `networking.k8s.io/v1/Ingress` instead which is present as of k8s v1.19 :+1:

iwilltry42:
@prometherion Thanks for the quick review! Updated :+1:

charts/capsule-proxy/templates/ingress.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "capsule-proxy.fullname" . -}}
-{{- if semverCompare ">=1.12-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" -}}
```
prometherion, line 3: Much smarter, loving it! 👏🏻

prometherion (approved):

PR #189, opened 2022-02-18
bug: cve exploiting malicious connection header

### LINKED ISSUE #188: Privilege escalation vulnerability via malicious "Connection" header
A user crafting an API request directed at `capsule-proxy` can get a privilege escalation using the Service Account of the proxy itself.

This is done by passing the `Impersonate-User` or `Impersonate-Group` header in the `Connection` header, using the same exploit described here: https://github.com/rancher/rancher/security/advisories/GHSA-pvxj-25m6-7vqr

> At this point, instead of impersonating the user and their permissions, the request will act as if it was from the ~~Rancher management server~~ Capsule Proxy and incorrectly return the information.

prometherion:
Closes #188.

We strongly suggest upgrading ASAP `capsule-proxy` to mitigate this CVE that doesn't have direct mitigation.

### DIFF
--- a/internal/webserver/webserver.go
+++ b/internal/webserver/webserver.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httputil"
+	"net/textproto"
 	"strings"
 	"time"
 
@@ -18,6 +19,7 @@ import (
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
+	"golang.org/x/net/http/httpguts"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apiserver/pkg/authentication/serviceaccount"
@@ -178,6 +180,9 @@ func (n kubeFilter) impersonateHandler(writer http.ResponseWriter, request *http
 	if len(n.bearerToken) > 0 {
 		request.Header.Set("Authorization", fmt.Sprintf("Bearer %s", n.bearerToken))
 	}
+	// Dropping malicious header connection
+	// https://github.com/clastix/capsule-proxy/issues/188
+	n.removingHopByHopHeaders(request)
 
 	request.Header.Add("Impersonate-User", username)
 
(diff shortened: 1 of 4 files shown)


internal/webserver/webserver.go
```diff
@@ -355,3 +358,27 @@ func (n kubeFilter) getProxyTenantsForOwnerKind(ownerKind capsulev1beta1.OwnerKi
...
+
+	connHeaders := request.Header.Values(headerName)
+
+	var sanitized []string
+
+	for _, values := range connHeaders {
+		parts := strings.Split(values, ",")
+
+		for _, part := range parts {
+			if strings.HasPrefix(strings.ToLower(part), "impersonate-") {
```
enj, line 373: Instead of removing bad headers, I would suggest following the approach taken by https://github.com/traefik/traefik/pull/8319 to resolve https://github.com/traefik/traefik/security/advisories/GHSA-m697-4v8f-55qg - i.e. handle the connection header directly.

With `if strings.HasPrefix(strings.ToLower(part), "impersonate-") {` you miss edge cases such as (note the leading spaces):

```
curl "https://${API}/api/v1/secrets" -H "Authorization: Bearer ${TOKEN}" -H "Connection: foo, Impersonate-User, Impersonate-Group" -0
```

enj:
@prometherion it would be great if a [GHSA](https://github.com/advisories) could be created once this fix is released.

internal/webserver/webserver.go
prometherion: Thanks for the hint!

internal/webserver/webserver.go
```diff
@@ -355,3 +360,28 @@ func (n kubeFilter) getProxyTenantsForOwnerKind(ownerKind capsulev1beta1.OwnerKi
 
 	return
 }
+
+func (n *kubeFilter) removingHopByHopHeaders(request *http.Request) {
+	connectionHeaderName, upgradeHeaderName, requestUpgradeType := "connection", "upgrade", ""
```
enj, line 365: These are case sensitive when used for direct header map access.

```suggestion
	connectionHeaderName, upgradeHeaderName, requestUpgradeType := "Connection", "Upgrade", ""
```

MaxFedotov: Or use 'request.Header.Values[connectionHeaderName]'

prometherion: Solved using the case-insensitive values retrieval function.

bsctl (approved):
lgtm thanks for the nice work

MaxFedotov (approved):
Lgtm!

PR #192, opened 2022-02-22
fix: retrieving connection values using case insensitive function

prometherion:


enj:
> @enj please, could you take a review of this PR?

Seems fine.  You should add tests 😄

Issue #194, opened 2022-03-01
Bearer token regex breaking ability to authenticate


jessecorson:
In `v0.2.0` and `v0.2.1` the regex check that uses `regexPatternForAuthHeader = "^(Bearer ([\\w-]*\\.[\\w-]*\\.[\\w-]*))$"` fails because our tokens do not adhere to that regex pattern. What is the purpose of this regex check?

It could still check for Bearer and a string, why does it need to be a jwt?

Reference https://github.com/clastix/capsule-proxy/blob/d85232c51624817909af86825fbc70c32e27be0e/internal/webserver/middleware/authorization.go#L19

prometherion:
Hey, thanks for using Capsule and Capsule Proxy!

That code has been written by @mendrugory, would be better ask him directly.

At first sight, it seems we have missed the support for web tokens since they were introduced with #99.

jessecorson:
@prometherion I've added the PR above to address this

mendrugory:
> Hey, thanks for using Capsule and Capsule Proxy!
> 
> That code has been written by @mendrugory, would be better ask him directly.
> 
> At first sight, it seems we have missed the support for web tokens since they were introduced with #99.

Thanks @jessecorson for pointing at this issue and for the proposed #195 .
That was lost after the big refactoring and only JWT were supported.

@prometherion should capsule-proxy support other kind of tokens?

prometherion:
> @prometherion should capsule-proxy support other kind of tokens?

Yes, we have, since supported by Kubernetes. We definitely lost it during refactoring!

PR #195, opened 2022-03-04
Fix webtoken auth

jessecorson:
Issue: the regex pattern for the checkBearerToken function currently does not support webtokens.

Solution: This PR added makes it so that both webtokens and jwts pass the regex

Reference https://github.com/clastix/capsule-proxy/issues/194


prometherion (approved):
Thanks, LGTM: let's check for the CI!

prometherion (changes requested):
And `golangci-lint` strikes again! Could you please address the issues?

jessecorson:
@prometherion hopefully those have been addressed

Issue #193, opened 2022-02-23
Refactoring tests


prometherion:
Tests are actually out of control.

Although `bats` is easy peasy, we're facing several issues with it, mostly regarding the copy/paste required across several test cases.

Furthermore, debugging the suite is tricky and error-prone, due to the limitations of bash.

We should investigate how to improve it, covering all the functionalities in each scenario (`kubectl`, http`, `TLS`, `User`, `Group`, `ServiceAccount`, etc.) and covering also #189 to avoid regressions.

oliverbaehler:
I will close this issue in favor of #257

PR #198, opened 2022-04-04
chore(tests): add tests for proxysettings custom resource

arriqaaq:
Adds tests for https://github.com/clastix/capsule-proxy/issues/197


arriqaaq:
@prometherion @MaxFedotov Should I replicate the proxysetting CR creation process for the other tests (like namespaces, nodes, priority classes, storage classes etc)? Since there is also support for owners through the tenant spec itself, and I assume we'd like to keep both type of tests (adding perms through tenant spec or through proxysetting CR), what are your thoughts?

prometherion:
I don't think we need to create so many cases, also considering #193, BATS is cool but actually, we're ending up with too many duplications of code.

Regarding the test case, since the feature is addressed as a middleware in the function `getTenantsForOwner`, there's no need to test the `ProxySetting` for each supported kind.

prometherion (approved):

prometherion:
Thanks a lot, @arriqaaq, going to merge tests on the base feature branch and fix CI to see the result of the changes! 👀

arriqaaq:
Thank you @prometherion , loved reading the code

PR #204, opened 2022-05-09
feat: support for owned namespace retrieval

### LINKED ISSUE #203: `kubectl apply` does not work for namespaces
With capsule proxy, when we try to create a namespace with `kubectl apply` it does not work and throws `403 Forbidden` error.  it does not work as the kubectl first queries the namespace, and that call results in the authorization error.

Steps to produce.

 namespace.yaml manifest file
```
apiVersion: v1
kind: Namespace
metadata:
 name: tnt-yaml
```
Applying the yaml definition
```
$ Kubectl apply -f namespace.yaml 
```

Error from server (Forbidden): error when retrieving current configuration of:
Resource: "/v1, Resource=namespaces", GroupVersionKind: "/v1, Kind=Namespace"
Name: "tnt-yaml", Namespace: ""
from server for: "namespace.yaml": namespaces "tnt-yaml" is forbidden: User "user1@gmail.com" cannot get resource "namespaces" in API group "" in the namespace "tnt-yaml"

prometherion:
Closes #203.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ kind:
 		&& kubectl taint nodes capsule-worker2 key1=value1:NoSchedule
 	@helm repo add bitnami https://charts.bitnami.com/bitnami
 	@helm upgrade --install --namespace metrics-system --create-namespace metrics-server bitnami/metrics-server \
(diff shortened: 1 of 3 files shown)


arriqaaq:
lgtm

PR #207, opened 2022-05-24
feat: expose HTTP requests metrics

viveksyngh:


prometherion:
Hey, this is astonishing! 💥 

Could you share a brief example of the exposed metrics here in the GitHub PR comments section? 👀

viveksyngh:
It exposes three metrics as the moment 

1. `capsule_proxy_requests_total` Total number of proxy requests 
2. `capsule_proxy_response_status` Number of requests by HTTP Status 
3.  `capsule_proxy_response_time_seconds` Proxy requests response time in seconds 

If there are additional metrics required or these metrics need to be changed, we can include them too. Please let me know.

prometherion (changes requested):
I see `golangci-lint` complaining about some issues, could try to fix them?

Otherwise, let's start the discussion on which linter should we disable.

viveksyngh:
@prometherion I have commented out `wsl` and `gofumpt` linters. I thought changes suggested by them did not make sense. 

I have also fixed other issues as well rather than having three metrics I have reduced them to two metrics. 
First two metrics were exactly the same but one with extra-label than other which I have combined in one. 

can I please get a review ?

internal/webserver/middleware/metrics.go
```diff
@@ -0,0 +1,63 @@
+package middleware
```
prometherion, line 1: Missing header here:m and on all the new files

```
// Copyright 2022 Clastix Labs
// SPDX-License-Identifier: Apache-2.0
```

.golangci.yml
```diff
@@ -32,6 +32,8 @@ linters:
     - goerr113
     - exhaustivestruct
     - lll
+    - wsl
+    - gofumpt
```
prometherion, line 36: What's the issue with these linters? They're pretty strict, reason is to enforce a coding style that is constant across all the files.

prometherion (changes requested):

prometherion (approved):

Issue #213, opened 2022-06-16
Helm: Add support for cert manager to generate SSL certificates 


viveksyngh:
Helm: Add support for cert-manager to generate SSL certificates

PR #214, opened 2022-06-16
feat: use cert-manager for SSL certificates

viveksyngh:
fixes: #213
Signed-off-by: Vivek Singh <svivekkumar@vmware.com>

### DIFF
--- a/charts/capsule-proxy/templates/certmanager.yaml
+++ b/charts/capsule-proxy/templates/certmanager.yaml
@@ -1,0 +1,52 @@
+{{- if and .Values.certManager.generateCertificates .Values.options.enableSSL -}}
+  {{- if not .Values.certManager.externalCA.enabled  -}}
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: {{ include "capsule-proxy.fullname" . }}-selfsigned-issuer
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: {{ include "capsule-proxy.fullname" . }}-selfsigned-ca
+spec:
+  isCA: true
+  commonName: {{ include "capsule-proxy.fullname" . }}-selfsigned-ca
+  secretName: {{ include "capsule-proxy.caSecretName" . }}
+  privateKey:
+    algorithm: ECDSA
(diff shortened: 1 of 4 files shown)


prometherion (changes requested):
Please, bump up the chart in order to get it published.

We have this condition and we have to avoid overwriting the previous version.

https://github.com/clastix/capsule-proxy/blob/7e711747a93d664336a562aa25c5f84f1362be39/.github/workflows/helm.yaml#L20

viveksyngh:
Updated the chart version to `0.2.5`

charts/capsule-proxy/templates/certmanager.yaml
```diff
@@ -0,0 +1,50 @@
...
+    kind: Issuer
+    group: cert-manager.io
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: {{ include "capsule-proxy.fullname" . }}-ca-issuer
+spec:
+  ca:
+    secretName: {{ include "capsule-proxy.fullname" . }}-root-secret
```
carpenterm, line 31: There are use cases where we may want to use an existing CA - say for example, the Kubernetes CA. Could we add some flag to enable/disable the self-signed CA certs and allow the user to provide the secret name used in the CA Issuer via the chart values i.e. the one at `spec.ca.secretName` above?

E.g.

```
certManager:
  generateCertificates: true
  existingCA:
    enabled: true
    secretName: my-custom-ca-cert
```

carpenterm: Thinking about it, `externalCA` may be a better name than `existingCA`. Just to add, the `certManager.externalCA.secretName` (or whatever we call it) would need to point to a secret containing the CA cert and private key of the external CA in the correct cert-manager format as per https://cert-manager.io/docs/configuration/ca/#deployment

prometherion: A helper could be implemented, something as the following:

```
{{- define "capsule-proxy.caSecretName" }}
{{- if .Values.ca.secretName }}
{{- printf "%s" .Values.ca.secretName }}
{{- else }}
{{- printf "%s-root-secret" (include "capsule-proxy.fullname" .) }}
{{- end }}
```

PR #219, opened 2022-06-18
feat(internal/request/http.go): check impersonation also of groups

maxgio92:
Signed-off-by: Massimiliano Giovagnoli <me@maxgio.it>

### DIFF
--- a/internal/request/http.go
+++ b/internal/request/http.go
@@ -12,6 +12,7 @@ import (
 	"github.com/golang-jwt/jwt"
 	authenticationv1 "k8s.io/api/authentication/v1"
 	authorizationv1 "k8s.io/api/authorization/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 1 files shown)


prometherion (commented):
I'm not aware of RBAC regarding impersonating groups in Kubernetes: could you point to the documentation about it?

Just saying that I always saw the verb `impersonate` about users, and not groups.

internal/request/http.go
```diff
@@ -176,3 +204,12 @@ func (h http) isJwtToken() bool {
 
 	return err == nil
 }
+
+func contains(ss []string, s string) bool {
```
prometherion, line 208: You can use the `sets.NewString` utility to check if an array contains a value. It's the same logarithmic complexity, it's much more readable.

prometherion (changes requested):

maxgio92:
> I'm not aware of RBAC regarding impersonating groups in Kubernetes: could you point to the documentation about it?
> 
> Just saying that I always saw the verb `impersonate` about users, and not groups.

Yes, [here](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation). AFAIU a `ClusterRole` rule on `impersonate` verb can apply to `["users", "groups", "serviceaccounts"]` resources.

internal/request/http.go
maxgio92: Applied @prometherion, thanks for pointing me out it

Issue #224, opened 2022-08-04
Exec/port-forward/logs command not working with capsule-proxy


viveksyngh:
Looks like exec/port-forward/logs command are not running with the latest version of capsule proxy

```
KUBECONFIG=alice-solar.capsule kubectl -n solar-development exec -it nginx-798bbb699b-rxfrk bash
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
Error from server: 
```

Capsule Proxy Logs
```
{"level":"Level(-4)","ts":"2022-08-04T10:14:18.942Z","logger":"proxy","msg":"impersonating for the current request","username":"alice","groups":["capsule.clastix.io"]}
{"level":"Level(-4)","ts":"2022-08-04T10:14:18.983Z","logger":"proxy","msg":"impersonating for the current request","username":"alice","groups":["capsule.clastix.io"]}
2022/08/04 10:14:18 http: proxy error: can't switch protocols using non-Hijacker ResponseWriter type *middleware.httpResponseWriter
```

Looks like the new middleware added for metrics collection is causing this issue. We are going to disable that middleware for now to fix this issue

PR #225, opened 2022-08-04
disable metrics middleware

viveksyngh:
Related to #224


prometherion:
Rather than a fix, I'd call this a workaround.

Merging this in order to provide a RC for the next patch release in order to unblock the WebSocket requests.

PR #231, opened 2022-09-01
regex for bearer token not valid for gcp access token

nvanrymenant:
The `CheckAuthorization` middleware in Capsule proxy contains a regex which checks of the bearer token is a valid JWT token.
GCP uses an access token for authentication which does not pass this regex check.

Steps to reproduce:

1. Made myself tenant owner using my GCP username:
```apiVersion: capsule.clastix.io/v1beta1
kind: Tenant
metadata:
  name: my-tenant
spec:
  additionalRoleBindings:
  - clusterRoleName: cluster-admin
    subjects:
    - name: gitops-reconciler
      kind: ServiceAccount
      namespace: klarriofc000policy
  owners:
  - name: system:serviceaccount:my-tenant:gitops-reconciler
    kind: ServiceAccount
  - name: nvanrymenant
    kind: User
```
2. Create local forward to capsule-proxy service: ` kubectl port-forward service/capsule-proxy 9001:9001 -n capsule-system`
3. Obtain GKE credentials: `gcloud container clusters get-credentials dp-gke-test --zone europe-west1`
4. Obtain id token: `/usr/lib/google-cloud-sdk/bin/gcloud config config-helper --format=json`
5. Query the Kubernetes API via Capsule proxy:
```
curl -kv --header "Authorization: Bearer ya29.A0AVAxxxxxxx" 'https://localhost:9001/api/v1/namespaces/klarriofc000policy-nginx'
```
Without the fix this request fails and results in `HandleUnauthorized` handler to be called.


prometherion (changes requested):
May I ask you to amend commit messages using the keywords according to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)?

nvanrymenant:
> May I ask you to amend commit messages using the keywords according to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)?

Rebased the PR, changed commit message and fixed an issue with the linter.

PR #237, opened 2022-11-08
feat: support ingressclassname in chart

carpenterm:
The ingress class annotation (e.g. `kubernetes.io/ingress.class: nginx`) has been deprecated since Kubernetes 1.18. This change adds support for setting `ingress.className` in the chart values which in turn sets the `spec.ingressClassName` field on the Ingress object for Capsule Proxy. `ingress.className` should refer to an existing `IngressClass` object. Note, the annotation and `ingressClassName` field cannot be used together or you'll get an error from the Kubernetes API.


bsctl (approved):
lgtm @carpenterm thanks for this contribution

bsctl (changes requested):
@carpenterm the lint is complaining, so please run make helm-docs locally and commit

bsctl (approved):

PR #243, opened 2022-12-05
fix: user with list permissions shouldn't be able to create objects

### LINKED ISSUE #242: User or group with LIST permission is able to create object
# Bug description

Tenant owner with `list` permissions on object is able to create them

# How to reproduce

create an owner with following permissions
```
  owners:
  - kind: Group
    name: foo
    proxySettings:
    - kind: IngressClasses
      operations:
      - List
    - kind: Nodes
      operations:
      - List
    - kind: StorageClasses
      operations:
      - List
```

User will be able to `create` ingress\storage class, although having only `list` permissions

# Expected behavior

An error should be returned to user

MaxFedotov:
closes #242


MaxFedotov:
> LGTM, please, just fix the `golangci-lint` errors!

done :)

PR #250, opened 2023-01-13
feat: support for clusterissuer in chart

sagar-jadhav:
Currently in [certmanager.yaml](https://github.com/clastix/capsule-proxy/blob/master/charts/capsule-proxy/templates/certmanager.yaml) following use cases are supported:

- Generate self-signed certificates for capsule-proxy using the self-signed issuer
- Generate the ca signed certificates for capsule-proxy using the ca-issuer

With this pull request I am adding support for one more use case details of which are as follows:

- I have a ClusterIssuer in the K8s cluster which is used by all the apps in the cluster for Certificate generation
- For capsule-proxy instead of using self signed issuer or ca issuer I want to use ClusterIssuer to generate capsule-proxy-serving-cert.

I refactored  [certmanager.yaml](https://github.com/clastix/capsule-proxy/blob/master/charts/capsule-proxy/templates/certmanager.yaml) to add support for this use case.

### DIFF
--- a/charts/capsule-proxy/Chart.yaml
+++ b/charts/capsule-proxy/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 0.3.2
 description: Helm Chart for Capsule Proxy, addon for Capsule, the multi-tenant Operator
 name: capsule-proxy
 type: application
-version: 0.3.3
+version: 0.3.4
 home: https://github.com/clastix/capsule-proxy
 icon: https://github.com/clastix/capsule/raw/master/assets/logo/capsule_small.png
 keywords:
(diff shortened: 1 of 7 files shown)


prometherion (changes requested):
LGTM, thanks for contributing! <3

May I ask you to bump-up the Chart version, so we can release it immediately after the merge?

sagar-jadhav:
> LGTM, thanks for contributing! <3
> 
> May I ask you to bump-up the Chart version, so we can release it immediately after the merge?

Thank you @prometherion, I have bump-up the chart version as requested.

prometherion (approved):

Issue #249, opened 2023-01-11
'unauthorized: forbidden access' error when deploying Capsule-Proxy on OpenShift 


rct44:
# Bug description
When capsule-proxy is deployed onto OpenShift and a user is authenticating to the cli via Bearer token, running an oc command via the proxy (e.g. oc get namespaces) returns an 'unauthorized: forbidden access' error. Due to the bearer token validation that happens within proxy, OpenShift tokens aren't being recognised and passed on when impersonating users, since they do not follow the same pattern as Kubernetes tokens. 

# How to reproduce
Deploy capsule-proxy to an OpenShift cluster. 
Configure the oc cli to use the proxy address as its server and a user authenticating with a bearer token
Run any oc command, e.g. oc get namespaces

# Expected behavior
'oc get namespaces' should return a list of namespaces of namespaces in tenant's owned by the user running the command

# Logs
Note that group names have been redacted and replaced with '\<group name>':
```
{"level":"Level(-4)","ts":"2023-01-09T13:46:13.346Z","logger":"proxy","msg":"allowed url path.","url path":"/api"}
{"level":"Level(-4)","ts":"2023-01-09T13:46:13.349Z","logger":"proxy","msg":"impersonating for the current request","username":"rachel","groups":["clastix-capsule","<group name>","<group name>"]}
{"level":"Level(-5)","ts":"2023-01-09T13:46:13.349Z","logger":"proxy","msg":"debugging request","uri":"/api?timeout=32s","method":"GET"}
{"level":"Level(-4)","ts":"2023-01-09T13:46:13.362Z","logger":"proxy","msg":"allowed url path.","url path":"/apis"}
{"level":"Level(-4)","ts":"2023-01-09T13:46:13.364Z","logger":"proxy","msg":"impersonating for the current request","username":"rachel","groups":["clastix-capsule","<group name>","<group name>"]}
{"level":"Level(-5)","ts":"2023-01-09T13:46:13.364Z","logger":"proxy","msg":"debugging request","uri":"/apis?timeout=32s","method":"GET"}
2023/01/09 13:46:13 unauthorized: forbidden access
2023/01/09 13:46:13 unauthorized: forbidden access
2023/01/09 13:46:13 unauthorized: forbidden access
```

# Additional context

- Capsule-Proxy version: 0.3.2
- Helm Chart version: 0.3.3
- OpenShift version: 4.10

Issue #245, opened 2022-12-30
Group Impersenation Handling


oliverbaehler:
# Bug description

Not sure if its considered a bug, but here's the current situation:

We are trying to implement [Teleport](https://goteleport.com/) to manage the access to our clusters. The Teleport workflow is as follows: From the terminal server the request is forwarded to a teleport agent installed on the target Kubernetes cluster. The Agent then uses the given credentials from the Terminal server and impersonates further requests with given user/groups ([Details](https://goteleport.com/docs/kubernetes-access/getting-started/#deployment-overview)). No we obviously directly want to connect to the capsule-proxy for a nice user experience for our users.  To achieve that we directed the teleport agent to the capsule-proxy endpoint :

```
extraEnv:
        - name: KUBERNETES_SERVICE_PORT_HTTPS
          value: "9001"
        - name: KUBERNETES_SERVICE_PORT
          value: "9001"
        - name: KUBERNETES_PORT_9001_TCP
          value: "tcp://capsule-proxy.capsule-system.svc:9001"
        - name: KUBERNETES_PORT_9001_TCP_PROTO
          value: "tcp"
        - name: KUBERNETES_PORT_9001_TCP_ADDR
          value: "capsule-proxy.capsule-system.svc"
        - name: KUBERNETES_SERVICE_HOST
          value: "capsule-proxy.capsule-system.svc"
        - name: KUBERNETES_PORT
          value: "tcp://capsule-proxy.capsule-system.svc:9001"
        - name: KUBERNETES_PORT_9001_TCP_PORT
          value: "443"
extraVolumeMounts:
  - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
    name: kube-api-access-capsule
    readOnly: true
extraVolumes:
  - name: kube-api-access-capsule
    projected:
      sources:
        - serviceAccountToken:
            path: token 
        - configMap:
            items:
              - key: ca.crt
                path: ca.crt
            name: capsule-root-ca.crt
        - downwardAPI:
            items:
              - fieldRef:
                  apiVersion: v1
                  fieldPath: metadata.namespace
                path: namespace
log:
        level: DEBUG


```
(Values for [teleport agent chart](https://github.com/gravitational/teleport/blob/master/examples/chart/teleport-kube-agent/README.md))

Great. That works. But now we are getting with every request the following error:

```
cannot retrieve user and group: the current user Oliver.Baehler@bedag.ch cannot impersonate the group company-group
```

[...]

Issue #57, opened 2021-03-11
Evaluate to switch generating local Table resources rather than reverse proxying to API Server


prometherion:
One key point of `capsule-proxy` is that we're just leveraging the Kubernetes API labels filtering capabilities to serve the expected resources a Tenant Owner (TO) is eager to get.

1. TO issues `kubectl get namespaces`
2. [`capsule-proxy` receives the request](https://github.com/clastix/capsule-proxy/blob/c34b1aa822c3e0fb958d76bab5d3b4466e3ed533/internal/webserver/webserver.go#L233-L236)
3. [decorates it with the additional label selector (`capsue.clastix.io/tenant in (oil,gas,...)`)](https://github.com/clastix/capsule-proxy/blob/c34b1aa822c3e0fb958d76bab5d3b4466e3ed533/internal/webserver/webserver.go#L190-L212)
4. [reverse proxy to the real API Server](https://github.com/clastix/capsule-proxy/blob/c34b1aa822c3e0fb958d76bab5d3b4466e3ed533/internal/webserver/webserver.go#L280-L297)
5. returning the response

This is pretty straightforward, no ticket science and just smart thinking.

However, with newer features request as the listing of Ingress and Storage classes, we need to hack a bit the resources since [fields selector doesn't support the `In` operator](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#supported-operators), forcing the Cluster Administrator to label the resources using the pattern `name=<resource-name>`.

What we could is, rather, serving directly the expected resources from `capsule-proxy`, without the need of doing the reverse proxy except for the case when this is needed (e.g.: when `capsule-proxy` is in front of the API Server as an adapter, collecting all the requests).

We got three kinds of resources we have to serve:
1. JSON output
2. YAML output (although converted from the kubectl binary(
3. Table output

Besides the kind that can be easily guessed by header `Accept` (e.g.: `Accept: application/json;as=Table;v=v1;g=meta.k8s.io,application/json;as=Table;v=v1beta1;g=meta.k8s.io,application/json`), we should support also the watch feature (`kubectl get resource --watch`), implemented with websockets.

## Table

I had the chance to dig the code-base of API Server and I was able to generate a JSON for the Table struct.

```go
package main

import (
	"context"

	appsv1 "k8s.io/api/apps/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/runtime"
[...]

PR #253, opened 2023-01-29
Tenant v1beta2 support

### LINKED ISSUE #246: Align with capsule v0.2.0 release
# Describe the feature

Align used API Versions with new versions from https://github.com/clastix/capsule/milestone/4. Make sure we release a new version of capsule proxy with allows a smooth migration to capsule 0.2.0 including migrating capsule-proxy.

Identified following changes:

- https://github.com/clastix/capsule-proxy/blob/master/main.go#L38: Add `capsulev1beta2` to scheme
-  https://github.com/clastix/capsule-proxy/blob/master/internal/controllers/capsule_configuration.go#L9: must be bumped to   `capsulev1beta2`
- https://github.com/clastix/capsule-proxy/blob/master/internal/controllers/role_bindings.go#L12: must be bumped to   `capsulev1beta2`

Test changes with main branch of capsule and verify everything is working as expected.

oliverbaehler:
Closes #246 

Bumps capsule to 0.2.0 and implements v1beta2. Will require a new major version as well. I would suggest releasing with the currently open features:

 * #249 
 * #245 

The feature for the selectors is only currently implemented for PriorityClasses. Before I continue I would like your feedback @prometherion. It would probably be better regarding performance to have the selector already preavaluated, than looping through all the available selectors here:

https://github.com/clastix/capsule-proxy/blob/issue/246/internal/modules/priorityclass/utils.go#L87

However I haven't found a way to do it. I noticed there's a `labels.Merge()` func, but this doesn't help. Maybe you have a smart way to return one selector object, which consists of all the proxyTenants selectors:

https://github.com/clastix/capsule-proxy/blob/issue/246/internal/modules/priorityclass/utils.go#L21


prometherion:
I think we should change the signature of `getPriorityClassSelector` in order to return a set of `*labels.Requirement` that will be used to create the Selector here:

https://github.com/clastix/capsule-proxy/blob/04006c61c5bc1a0e391eb151ec9eec46773abc5f/internal/modules/priorityclass/list.go#L61

I'll try to work on a bit on your changes.

prometherion:
@oliverbaehler take a look [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements)

> Caution: For both equality-based and set-based conditions there is no logical OR (||) operator. Ensure your filter statements are structured accordingly.

With that said, we need to make `capsule-proxy` a bit hacky, I'd say: `matchLabels` and `matchRequirements` will have preference over the `regex` and `name`, also considering we plan to deprecate the latter ones since not properly aligned to the set-base selection philosophy in Kubernetes.

WDYT?

oliverbaehler:
@prometherion annoying that OR is not supported. We are facing the same dilemma with the Default tenant value. I guess would only be solvable with something like #57

prometherion:
Table generation is unfeasible, at least in the short term.

Let's go for prioritizing the matchLabels/matchRequirements over the exact/regex matches, also considering that the latter ones will be deprecated in future by Capsule.

oliverbaehler:
@prometherion What about the default? should we just not support it for now?

prometherion:
We have to inform end-users that even the default must match the matchLabels/matchRequirements, although not required in Capsule, just for listing purposes.

prometherion:
Closes #266 too with commit 34eba3b.

PR #255, opened 2023-02-03
fix: unable to list nodes as a tenant owner (#254)

### LINKED ISSUE #254: unable to list nodes as a tenant owner
# Bug description

As a tenant owner I am not able to list the nodes assigned to me.

# How to reproduce

As a cluster admin label the node as follows

```
kubectl label node localhost org=oil
```

As a cluster admin create the following tenant object:

```
kind: Tenant
metadata:
  name: oil
spec:
  owners:
    - kind: User
      name: alice
      proxySettings:
        - kind: Nodes
          operations:
            - List
  nodeSelector:
    org: oil
```

Login to the k8s cluster using tenant owner kubeconfig and run the following command:

```
kubectl get nodes

No resource found
```
# Expected behavior

As tenant owner I should be able to list the nodes with label `org: oil`

# Additional context

- Capsule-Proxy version: v0.3.2
- Helm Chart version: capsule-proxy-0.3.1
- Kubernetes version: Client Version: v1.23.16, Server Version: v1.22.7

# Notes

 I debugged the issue and found that [PR-239](https://github.com/clastix/capsule-proxy/pull/239) introduced this bug. Below pattern is incorrect:

```
"/api/v1/{nodes:[^/]+/?}"
```

Here mux will only capture the value of the "nodes" parameter as the portion of the URL that matches the pattern `[^/]+/?` So nodes is the name of the parameter and not the nodes as string. `[^/]+` signifies any number of character other than `/` and `/?` signifies that zero or one `/` in the URL. 

So when [mod.Path()](https://github.com/clastix/capsule-proxy/blob/501f017b4571ddbee9881e377d046c8ac5ffc18f/internal/webserver/webserver.go#L228) method is invoked for namespace module with path **/api/v1/{namespaces:[^/]+/?}** a route is registered which matches with the below pattern

```
/api/v1/somecharacters
```

for e.g. /api/v1/nodes, /api/v1/namespace etc

So all calls starting with `/api/v1/somecharacters` pattern is handled by the namespace handler because of **/api/v1/{namespaces:[^/]+/?}** pattern. 

I changed the pattern from `/api/v1/{namespaces:[^/]+/?}` to `/api/v1/{endpoint:namespaces/?}` and able to resolve the issue. Here endpoint is name of the parameter and `namespaces/?` signifies that word should start with namespaces string and should or should not end with `/`

sagar-jadhav:
Fixes #254

### DIFF
--- a/internal/modules/ingressclass/list.go
+++ b/internal/modules/ingressclass/list.go
@@ -31,7 +31,7 @@ func List(client client.Client) modules.Module {
 }
 
 func (l list) Path() string {
-	return "/apis/networking.k8s.io/{version}/{ingressclasses:[^/]+/?}"
+	return "/apis/networking.k8s.io/{version}/{endpoint:ingressclasses/?}"
 }
 
 func (l list) Methods() []string {
@@ -39,7 +39,7 @@ func (l list) Methods() []string {
 }
 
 func (l list) Subrouter(router *mux.Router) *mux.Router {
-	return router.Path("/apis/networking.k8s.io/{version}/{ingressclasses:[^/]+/?}").Subrouter()
+	return router.Path("/apis/networking.k8s.io/{version}/{endpoint:ingressclasses/?}").Subrouter()
 }
 
 func (l list) Handle(proxyTenants []*tenant.ProxyTenant, proxyRequest request.Request) (selector labels.Selector, err error) {
--- a/internal/modules/metric/list.go
+++ b/internal/modules/metric/list.go
@@ -31,7 +31,7 @@ func List(client client.Client) modules.Module {
 }
 
 func (l list) Path() string {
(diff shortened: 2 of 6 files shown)


prometherion:
Tested locally the patch:

```
$: curl --cert alice-oil.crt --key alice-oil.key https://127.0.0.1:9001/api/v1/nodes/
$: curl --cert alice-oil.crt --key alice-oil.key https://127.0.0.1:9001/api/v1/nodes
```
In both cases, I have the node lists belonging to the tenant.

@MaxFedotov LGTM, may I ask you for a review, please?

prometherion (approved):

MaxFedotov:
@prometherion tried it locally to - everything is working perfectly, so lgtm to mee too.
@sagar-jadhav thanks a lot for this fix!

PR #256, opened 2023-02-03
fix(test): metrics-server chart bump up to v6.2.9

prometherion:
It should fix the CI 🙏🏻


prometherion:
This PR is not going to solve the CI issue we're facing, issue #257 is tracking this enhancement, these changes would like to fix just the local setup that should (famous last words) work.

PR #259, opened 2023-02-09
feat: ability to handle multiple auth types

### LINKED ISSUE #258: Handle Multiple Authentication Types
We currently are attempting to expose capsule-proxy via [Contour](https://projectcontour.io). Contour is configured to send a client cert with each request proxied to a backend service to allow the backend service to validate that the request is coming from an expected source, which capsule-proxy has support for. However the cert is supplied by capsule and not by the client so there are no username/group claims present on the cert.

I'd like to be able to somehow specify which authentication type we are wanting to use, or be able to configure the order in which types are tried [here](https://github.com/clastix/capsule-proxy/blob/v0.3.2/internal/request/http.go#L138-L147). 

I can do the work just looking for some feedback on what the best way to accomplish this would be.

Thanks!

logikone:
closes #258


prometherion:
@logikone just noticed some minor things with `golangci-lint`, fixed and force-pushed.

e2e is broken and we're working to fix it, I want to focus on minor lint things, and get it merged.

main.go
```diff
@@ -60,6 +62,16 @@ func main() {
...
 	var rolebindingsResyncPeriod time.Duration
 
+	authTypes := []request.AuthType{
+		request.TLSCertificate,
+		request.BearerToken,
+	}
+
+	authTypesMap := map[request.AuthType][]string{
+		request.BearerToken:    {request.BearerToken.String(), "bearer", "token"},
+		request.TLSCertificate: {request.TLSCertificate.String(), "certificate", "cert"},
```
prometherion, line 72: Do we really need multiple values for these enums? It could create confusion and I would stick to the `String` representation.

main.go
```diff
@@ -69,6 +81,10 @@ func main() {
 	flag.StringVar(&certPath, "ssl-cert-path", "", "Path to the TLS certificate (default: /opt/capsule-proxy/tls.crt)")
 	flag.StringVar(&keyPath, "ssl-key-path", "", "Path to the TLS certificate key (default: /opt/capsule-proxy/tls.key)")
 	flag.DurationVar(&rolebindingsResyncPeriod, "rolebindings-resync-period", 10*time.Hour, "Resync period for rolebindings reflector")
+	flag.Var(enumflag.NewSlice(&authTypes, "string", authTypesMap, enumflag.EnumCaseInsensitive), "auth-preferred-types",
```
prometherion, line 84: I would skip the `enumflag.EnumCaseInsensitive`, since it's a human input, we just need to document it properly.

prometherion (commented):
Some minor remarks, let's discuss them.

main.go
logikone: ok, i've reduce these lists to just the string repr for each

main.go
logikone: I've changed the enum flag to be case sensitive and have updated the flags docstring

prometherion (approved):
Force pushed some changes due to `golangci-lint` complaints, LGTM, thanks a lot @logikone! 🎉

logikone:
you're welcome! thank you!

PR #269, opened 2023-03-08
fix: do not expose service name to anonymous requests

slimm609:
If you are not authenticated, don't return the `capsule` name in the error message. This helps reduce information leaking to anonymous requests.


prometherion (approved):
LGTM although CI is broken, thanks!

PR #270, opened 2023-03-08
remove symbols table

slimm609:
- remove symbols table
- trimpath of absolute compile path

This removes debug info from the final binary and reduces the size by about 25%. These flags won't be included during the dlv-build so should have no impact on debugging build.

on arm64 reduced size by almost 12MB.
49.1MB -> 37.4MB


prometherion (changes requested):
Thanks, definitely needed!

May I ask you to reword the subject to adhere to the semantic commit messages?

slimm609:
updated in both the PR and the commit message itself. Also updated lint version

slimm609:
should be good now. This is a documented bug in the action, updating to 3.0 no longer installs go but since it's installed right before it's not a problem and fixes the linter fmt missing issue.

PR #272, opened 2023-03-09
only verify client cert if TLSCertAuth is active

### LINKED ISSUE #271: don't validate client cert if TLSCertificate is not active
If the `--auth-preferred-types` cli flag is provided and `TLSCertificate` is not listed then there is no need validate a certificate if one is sent.

We're having troubles with this currently since the ingress controller we're using to expose capsule-proxy sends a client cert that is not signed by the kubernetes ca.

logikone:
* If `TLSCertAuth` is not provided as one of the preferred auth types then skip verification

closes #271

### DIFF
--- a/internal/webserver/webserver.go
+++ b/internal/webserver/webserver.go
@@ -275,6 +275,7 @@ func (n *kubeFilter) registerModules(ctx context.Context, root *mux.Router) {
 	}
 }
 
+//nolint:funlen
 func (n *kubeFilter) Start(ctx context.Context) error {
 	r := mux.NewRouter()
 	r.Use(handlers.RecoveryHandler())
@@ -306,8 +307,16 @@ func (n *kubeFilter) Start(ctx context.Context) error {
 			tlsConfig := &tls.Config{
 				MinVersion: tls.VersionTLS12,
 				ClientCAs:  n.serverOptions.GetCertificateAuthorityPool(),
-				ClientAuth: tls.VerifyClientCertIfGiven,
 			}
+
+			for _, authType := range n.authTypes {
+				if authType == req.TLSCertificate {
+					tlsConfig.ClientAuth = tls.VerifyClientCertIfGiven
+
+					break
+				}
+			}
+
 			srv = &http.Server{
 				Handler:   r,
 				Addr:      addr,


internal/webserver/webserver.go
```diff
@@ -306,8 +306,15 @@ func (n *kubeFilter) Start(ctx context.Context) error {
 			tlsConfig := &tls.Config{
 				MinVersion: tls.VersionTLS12,
 				ClientCAs:  n.serverOptions.GetCertificateAuthorityPool(),
-				ClientAuth: tls.VerifyClientCertIfGiven,
+				ClientAuth: tls.NoClientCert,
```
prometherion, line 309: `tls.NoClientCert` is the default value, according to the `stdlib` package.

```
	// ClientAuth determines the server's policy for
	// TLS Client Authentication. The default is NoClientCert.
	ClientAuth ClientAuthType
```

internal/webserver/webserver.go
```diff
@@ -306,8 +306,15 @@ func (n *kubeFilter) Start(ctx context.Context) error {
...
 			}
+
+			for _, authType := range n.authTypes {
+				if authType == req.TLSCertificate {
+					tlsConfig.ClientAuth = tls.VerifyClientCertIfGiven
```
prometherion, line 314: Once found, we should `break` the iteration.

prometherion (changes requested):

prometherion:
@MaxFedotov I would appreciate your help on this since I have some concerns, please, @logikone feel free to join the discussion!

Could mTLS disabling allow a privilege escalation in the proxy? This shouldn't happen, theoretically, since we got the function `GetUserAndGroups` that is returning the user according to the preferred authentication type.

internal/webserver/webserver.go
logikone: ok, i can remove this line 👍

internal/webserver/webserver.go
logikone: added break statement. thank you!

logikone:
I don't think it should allow privilege escalation if the cert is not being checked for claims. Though one thing I did notice is that, once I changed the client cert that our ingress controller is sending to one that was signed by the kube ca, every request was able to list all namespaces. The cert was present and valid but had no claims, and the bearer token was not checked at all. This seems like it might be a bug. My assumption was that if the cert did not have any claims the request should have no access.

This could have been something else strange going on with the way I had things setup at the time though, and I have not done any troubleshooting to try and repro.

prometherion:
> Though one thing I did notice is that, once I changed the client cert that our ingress controller is sending to one that was signed by the kube ca, every request was able to list all namespaces.

I'd say this happened because you were using the certificate of the Ingress Controller that has list capabilities of Namespace resources. If this isn't the case, a bug case should be rise to investigate it properly and address a fix.

prometherion (approved):
I had the chance to give it a try to your changes, added some minor fixes and everything seems good, thanks for the contribution! ❤️

PR #274, opened 2023-03-12
fix: fallback to multiple auth strategies

### LINKED ISSUE #273: --auth-preferred-types flag can be set multiple times, just first option is evaluated
# Bug description

A new CLI flag to select the preferred authentication mechanism has been introduced with #258 (`--auth-preferred-types`): this retrieves the username, and the groups, from the handled requests according to the authentication strategy provided.

However, only the first strategy is put in place, ignoring the following one(s).

# How to reproduce

1. Start the Capsule Proxy with the bearer token and TLS
2. Use the wrong token in order to get backed by the TLS certificate
3. Get the error of the token, noticing that the client certificate is not checked

# Expected behaviour

According to the ordered strategy, the fallback should be put in place.

# Logs

n.r.

# Additional context

- Capsule-Proxy version: (`v0.4.0,`)
- Helm Chart version: (`v0.4.0,`)
- Kubernetes version: (`n.r.`)

prometherion:
Closes #273.

### DIFF
--- a/internal/request/http.go
+++ b/internal/request/http.go
@@ -30,18 +30,12 @@ func (h http) GetHTTPRequest() *h.Request {
 
 //nolint:funlen
 func (h http) GetUserAndGroups() (username string, groups []string, err error) {
-	switch h.getAuthType() {
-	case TLSCertificate:
-		pc := h.TLS.PeerCertificates
-		if len(pc) == 0 {
-			return "", nil, fmt.Errorf("no provided peer certificates")
+	for _, fn := range h.authenticationFns() {
+		// User authentication data is extracted according to the preferred order:
+		// in case of first match blocking the iteration
+		if username, groups, err = fn(); err == nil {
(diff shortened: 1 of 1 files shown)


prometherion:
@logikone may I ask you for a brief review of this? The idea is to have a preferred fallback in case one of the strategies failed in retrieving the required auth info.

prometherion:
Tested locally and it works as expected when running the proxy as `--auth-preferred-types=TLSCertificate --auth-preferred-types=BearerToken`:

1. using a `kubeconfig` backed by a certificate and a token
2. when the cert is ok, auth data is extracted from the peer certificates
3. if the cert is not ok, the token is used
4. otherwise, falling back to the unauthenticated that stops the flow

logikone:
Ya the way it worked before my change was also not falling through to token auth if cert auth failed so this will be great! Thank you, sir!

PR #290, opened 2023-05-10
fix: update golang to 1.19

slimm609:
update golang to 1.19 to allow for easier recompile using boringcrypto added to main in 1.19. Also moved from alpine to just regular golang to standardize between capsule and capsule-proxy.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,4 +20,4 @@ jobs:
         with:
           version: v1.51.2
           only-new-issues: false
-          args: --timeout 2m --config .golangci.yml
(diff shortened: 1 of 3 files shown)


prometherion:
LGTM @slimm609 but I wouldn't define this as a fix, rather, `deps` as a semantic commit subject.

The GH changes could be marked as `chore(github)`.

Once amended commit messages, ready to get it merged.

slimm609:
rebased and amended into multiple

slimm609:
@prometherion this should be good now

prometherion (approved):
Thanks, @slimm609!

PR #293, opened 2023-05-13
fix: ensure capsuleconfiguration exists at startup

### LINKED ISSUE #292: missing start up arguments error handling
# Bug description

Capsule-proxy does not fail when a non existing `--capsule-configuration-name` is specified in its start up arguments.

# How to reproduce

Start capsule-proxy with a non existing CapsuleConfiguration CRD

`--capsule-configuration-name=blablabla`

# Expected behavior

Fail to start capsule-proxy because the specified CapsuleConfiguration CRD does not exist

prometherion:
Closes #292.

@latchmihay may I ask you for a review?

### DIFF
--- a/internal/controllers/capsule_configuration.go
+++ b/internal/controllers/capsule_configuration.go
@@ -5,8 +5,11 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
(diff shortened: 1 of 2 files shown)


latchmihay (approved):
Easy enough, LGTM!

PR #294, opened 2023-05-16
fix: add imagePullSecrets to certgen-job

kaotika:
The certgen-job does not use imagePullSecrets if they are set in Values.yaml. The PR fixes this.


kaotika:
@prometherion thanks for quick response. I bumped the chart version as requested.

PR #299, opened 2023-05-25
fix: imagePullSecrets belongs to spec instead of containers

NyQuilXT:
imagePullSecrets were misplaced under containers instead directly under spec. This PR will fix this.


prometherion (approved):
Thanks a lot, mate! 🤗

PR #300, opened 2023-05-26
feat: add client side rate limiters

abhinandanbaheti:
Currently, client side connections for Capsule proxy to API Server are throttling, and adding latency to users. There is no option to configure rate limiters - QPS and Burst. These configurable param can be added to address these latency and performance issues.  
Default Controller runtime rate limiters  are QPS (20.0), Burst (30), we can probably use them as default  
Ref - https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/config/config.go#L96-L108

Capsule proxy logs are filled with following log lines

I0524 19:59:09.232782       1 request.go:601] Waited for 7.652678944s due to client-side throttling, not priority and fairness, request: POST:[https://x.x.x.x:443/apis/authentication.k8s.io/v1/tokenreviews](https://100.64.0.1/apis/authentication.k8s.io/v1/tokenreviews)
I0524 19:59:19.279022       1 request.go:601] Waited for 5.4349592s due to client-side throttling, not priority and fairness, request: POST:[https://x.x.x.x:443/apis/authentication.k8s.io/v1/tokenreviews](https://100.64.0.1/apis/authentication.k8s.io/v1/tokenreviews)
I0524 19:59:56.739446       1 request.go:601] Waited for 1.011912346s due to client-side throttling, not priority and fairness, request: POST:[https://x.x.x.x:443/apis/authentication.k8s.io/v1/tokenreviews](https://100.64.0.1/apis/authentication.k8s.io/v1/tokenreviews)
I0524 20:00:06.739694       1 request.go:601] Waited for 8.498116916s due to client-side throttling, not priority and fairness, request: POST:[https://x.x.x.x:443/apis/authentication.k8s.io/v1/tokenreviews](https://100.64.0.1/apis/authentication.k8s.io/v1/tokenreviews)
I0524 20:02:02.691683       1 request.go:601] Waited for 1.044931468s due to client-side throttling, not priority and fairness, request: POST:[https://x.x.x.x:443/apis/authentication.k8s.io/v1/tokenreviews](https://100.64.0.1/apis/authentication.k8s.io/v1/tokenreviews)

PR #304, opened 2023-06-12
fix(helm): add client side rate limiters

abhinandanbaheti:
This is to add client side rate limiters in the helm manifests

Changes with code are already merged in the following PR's and released with v0.4.4.
https://github.com/clastix/capsule-proxy/pull/300
https://github.com/clastix/capsule-proxy/pull/301


prometherion (changes requested):
Let's bump up the Chart.yaml file so we can release it! 👍🏻

prometherion (approved):

PR #308, opened 2023-08-21
WIP: intercept anything feature

oliverbaehler:


prometherion:
@oliverbaehler I had some progress on this, such as testing if the proxy user has permission to retrieve the objects in the Tenant namespaces.

It would be great if we could give it a try in an extensive way, unfortunately, our e2e suite is totally broken ad useless.

prometherion:
Closing in favour of #389.

PR #313, opened 2023-10-05
feat(helm): add annotations for certgen job and make ttlSecondsAfterFinished optional

adberger:
This is needed if capsule-proxy is rolled out with ArgoCD, see: https://argo-cd.readthedocs.io/en/stable/user-guide/resource_hooks/#hook-deletion-policies

### DIFF
--- a/charts/capsule-proxy/Chart.yaml
+++ b/charts/capsule-proxy/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 0.4.4
 description: Helm Chart for Capsule Proxy, addon for Capsule, the multi-tenant Operator
 name: capsule-proxy
 type: application
-version: 0.4.5
+version: 0.5.0
 home: https://github.com/clastix/capsule-proxy
 icon: https://github.com/clastix/capsule/raw/master/assets/logo/capsule_small.png
 keywords:
--- a/charts/capsule-proxy/templates/certgen-job.yaml
+++ b/charts/capsule-proxy/templates/certgen-job.yaml
@@ -5,8 +5,14 @@ metadata:
   name: {{ include "capsule-proxy.fullname" . }}-certgen
   labels:
     {{- include "capsule-proxy.labels" . | nindent 4 }}
+  {{- with .Values.jobs.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
 spec:
-  ttlSecondsAfterFinished: 60
+  {{- if ge .Values.jobs.ttlSecondsAfterFinished 0.0 }}
+  ttlSecondsAfterFinished: {{ .Values.jobs.ttlSecondsAfterFinished }}
+  {{- end }}
   template:
     metadata:
       name: {{ include "capsule-proxy.fullname" . }}-certgen
--- a/charts/capsule-proxy/README.md
+++ b/charts/capsule-proxy/README.md
@@ -72,11 +72,13 @@ If you only need to make minor customizations, you can specify them on the comma
 | image.repository | string | `"clastix/capsule-proxy"` | Set the image repository of the capsule-proxy. |
 | image.tag | string | `""` | Overrides the image tag whose default is the chart appVersion. |
 | imagePullSecrets | list | `[]` | Configuration for `imagePullSecrets` so that you can use a private images registry. |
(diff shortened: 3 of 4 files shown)


prometherion:
Thanks for the contribution, @adberger!

charts/capsule-proxy/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: 0.4.4
 description: Helm Chart for Capsule Proxy, addon for Capsule, the multi-tenant Operator
 name: capsule-proxy
 type: application
-version: 0.4.5
+version: 0.5.0
```
prometherion, line 6: A minor bump here could be good, since we're adding new fields.

charts/capsule-proxy/templates/certgen-job.yaml
```diff
@@ -5,8 +5,14 @@ metadata:
...
   labels:
     {{- include "capsule-proxy.labels" . | nindent 4 }}
+  {{- with .Values.jobs.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
 spec:
+  {{- if .Values.jobs.deleteAfterFinished }}
   ttlSecondsAfterFinished: 60
+  {{- end }}
```
prometherion, line 15: Wouldn't be better to have this value configurable? Something has:

```
{{- if gt .Values.jobs.ttlSecondsAfterFinished 0 }}
  ttlSecondsAfterFinished: {{ .Values.jobs.ttlSecondsAfterFinished }}
{{- end }}
```

adberger: Can do this, yes =)

charts/capsule-proxy/Chart.yaml
adberger: Don't you use semVer2 for versioining? This is a minor bump.  
0.4.6 would be a bugfix.

charts/capsule-proxy/templates/certgen-job.yaml
adberger: @prometherion Can you take another look? Because `0` is also valid (delete immediately) I used `ge` instead of `gt`.  
For disabling I used `-1`.

charts/capsule-proxy/templates/certgen-job.yaml
```diff
@@ -5,8 +5,14 @@ metadata:
...
     {{- include "capsule-proxy.labels" . | nindent 4 }}
+  {{- with .Values.jobs.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
 spec:
-  ttlSecondsAfterFinished: 60
+  {{- if ge .Values.jobs.ttlSecondsAfterFinished 0.0 }}
+  ttlSecondsAfterFinished: {{ .Values.jobs.ttlSecondsAfterFinished }}
+  {{- end }}
```
prometherion, line 15: Are you sure `ge` is the right operator here?

Otherwise, we'll have the `ttlSecondsAfterFinished` value to zero, and according to the documentation:

```
If this field is set to zero, the Job becomes eligible to be deleted immediately after it finishes.
```

charts/capsule-proxy/templates/certgen-job.yaml
prometherion: Definitely makes sense, you're right, and it's already documented.

prometherion (approved):

prometherion:
Please, may I ask you to share the example values you're using to let it work with ArgoCD?

The idea is to share with the community, so in a glimpse they can found the solution to their problems.

charts/capsule-proxy/templates/certgen-job.yaml
adberger: Yeah, maybe somebody intends to do that?  
As of `0` is a valid value for the spec, I would make it possible to set it.

adberger:
> Please, may I ask you to share the example values you're using to let it work with ArgoCD?
> 
> The idea is to share with the community, so in a glimpse they can found the solution to their problems.

Sure:

A minimal `values.yaml` for working with ArgoCD could be:
```
jobs:
  annotations:
    argocd.argoproj.io/hook: PostSync
    argocd.argoproj.io/hook-delete-policy: HookSucceeded
  ttlSecondsAfterFinished: -1
```

PR #316, opened 2023-10-20
Moving to the neutral GitHub organization

prometherion:
As the CNCF sandbox donation continues, also the proxy must be moved.


oliverbaehler (approved):
LGTM

PR #327, opened 2023-10-31
fix: supporting unauthenticated version endpoint

### LINKED ISSUE #315: Kubernetes Dashboard crashing on startup with capsule-proxy
We have a runing setup of Kubernetes with Guard (AAD integration) and the official Kubernetes Dashboard. After login over hte oauth2-proxy we having a jwt token which will be used automatically in the dashboard.

As I'm triyng to implementd capsule in our setup I tried to use the capsule-proxy for the users to see their workload over the dashboard. So I've deployed the capsule-proxy with helm on a seprated ingress url. 

For the kubectl setup this is working fine and the users can only see their workload.
Starting to implement the capsule-proxy in the dashboard setup, I got some issues. 

I added the param `--apiserver-host=http://capsule-proxy.capsule-system:9001` into the dashboard args section but the dashboard cannot startup with the following erros: 

```2023/10/19 07:00:00 Using apiserver-host location: http://capsule-proxy.capsule-system:9001
2023/10/19 07:00:00 Using namespace: kubernetes-dashboard
2023/10/19 07:00:00 Skipping in-cluster config
2023/10/19 07:00:00 Using random key for csrf signing
2023/10/19 07:00:00 Successful initial request to the apiserver, version:
2023/10/19 07:00:00 Generating JWE encryption key
2023/10/19 07:00:00 New synchronizer has been registered: kubernetes-dashboard-key-holder-kubernetes-dashboard. Starting
2023/10/19 07:00:00 Starting secret synchronizer for kubernetes-dashboard-key-holder in namespace kubernetes-dashboard
2023/10/19 07:00:00 Synchronizer kubernetes-dashboard-key-holder-kubernetes-dashboard exited with error: unexpected object: &Secret{ObjectMeta:{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] [] []},Data:map[string][]byte{},Type:,StringData:map[string]string{},Immutable:nil,}
2023/10/19 07:00:00 Storing encryption key in a secret
panic: cannot retrieve user and group: unauthenticated users not supported
```
In the capsule-proxy log I can see the following:

```{"level":"Level(-4)","ts":"2023-10-19T07:00:00.075Z","logger":"proxy","msg":"allowed url path.","url path":"/version"}
2023/10/19 07:00:00 http: superfluous response.WriteHeader call from github.com/gorilla/handlers.recoveryHandler.ServeHTTP.func1 (recovery.go:73)
2023/10/19 07:00:00 cannot retrieve user and group: unauthenticated users not supported
2023/10/19 07:00:00 http: superfluous response.WriteHeader call from github.com/gorilla/handlers.recoveryHandler.ServeHTTP.func1 (recovery.go:73)
2023/10/19 07:00:00 cannot retrieve user and group: unauthenticated users not supported
2023/10/19 07:00:00 http: superfluous response.WriteHeader call from github.com/gorilla/handlers.recoveryHandler.ServeHTTP.func1 (recovery.go:73)
2023/10/19 07:00:00 cannot retrieve user and group: unauthenticated users not supported
2023/10/19 07:00:00 http: superfluous response.WriteHeader call from github.com/gorilla/handlers.recoveryHandler.ServeHTTP.func1 (recovery.go:73)
2023/10/19 07:00:00 cannot retrieve user and group: unauthenticated users not supported
```

So the dashboard tries to get `/version` from the API but is not allowed and crashed. Did I missed something in the setup? Because I don't understand why the dashboard does a query without any auth to the api and I can find any hints in your docu about that.

- Kubernetes Dashboard: 2.7.0 and 3.0.0-alpha
- Capsule version: 0.3.3
- Capsule-Proxy version: 0.4.4
- Kubernetes version: 1.26.6

prometherion:
Closes #315.

### DIFF
--- a/internal/webserver/webserver.go
+++ b/internal/webserver/webserver.go
@@ -296,6 +296,36 @@ func (n *kubeFilter) Start(ctx context.Context) error {
 
 	root := r.PathPrefix("").Subrouter()
 	n.registerModules(ctx, root)
+	// Allowing a non-authenticated proxy pass for the /version endpoint in case of anonymous auth.
+	// This is required since most of the Kubernetes distributions are exposing the version endpoint with not auth.
+	// If the proxy is used by an application, such as the Kubernetes dashboard, which is not adding an authentication
+	// strategy, this would fail since unable to extract the cluster version.
+	root.Path("/version").HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+		hr := req.NewHTTP(request, n.authTypes, n.usernameClaimField, n.writer)
+
+		username, groups, err := hr.GetUserAndGroups()
+		if err != nil {
+			var t *req.ErrUnauthorized
+			if !errors.As(err, &t) {
+				server.HandleError(writer, err, "")
+			}
+		}
+		// Dropping malicious header connection
+		// https://github.com/projectcapsule/capsule-proxy/issues/188
+		n.removingHopByHopHeaders(request)
+
+		if len(username) > 0 {
+			if len(n.bearerToken) > 0 {
+				request.Header.Set("Authorization", fmt.Sprintf("Bearer %s", n.bearerToken))
+			}
+
+			request.Header.Add(authenticationv1.ImpersonateUserHeader, username)
+
+			for _, group := range groups {
+				request.Header.Add(authenticationv1.ImpersonateGroupHeader, group)
+			}
+		}
+	})
 	root.Use(
 		n.reverseProxyMiddleware,
 		middleware.CheckPaths(n.log, n.allowedPaths, n.impersonateHandler),
--- a/internal/webserver/webserver.go
+++ b/internal/webserver/webserver.go
@@ -62,7 +62,7 @@ func NewKubeFilter(opts options.ListenerOpts, srv options.ServerOptions, rbRefle
 
 	return &kubeFilter{
 		reader:                clientOverride,
-		allowedPaths:          sets.NewString("/api", "/apis", "/version"),
+		allowedPaths:          sets.NewString("/api", "/apis"),
 		authTypes:             opts.AuthTypes(),
 		ignoredUserGroups:     sets.NewString(opts.IgnoredGroupNames()...),
 		reverseProxy:          reverseProxy,


prometherion:
@sybnex I tested this personally and seems fixing what you're looking for.

```
$: curl -k https://127.0.0.1:9001/version
{
  "major": "1",
  "minor": "25",
  "gitVersion": "v1.25.3",
  "gitCommit": "434bfd82814af038ad94d62ebe59b133fcb50506",
  "gitTreeState": "clean",
  "buildDate": "2022-10-25T19:35:11Z",
  "goVersion": "go1.19.2",
  "compiler": "gc",
  "platform": "linux/amd64"
}
```

For the `/version` endpoint we're skipping the error in case of retrieval of authentication data, bypassing the user impersonation and decorating the request with the bearer token which would be required to perform the necessary user impersonation.

I wanted to give it a try also if we support a protected endpoint, and it seems to do the trick:

```
$: curl -k -H 'Authorization: Bearer <REDACTED>' https://127.0.0.1:9001/version
{
  "major": "1",
  "minor": "25",
  "gitVersion": "v1.25.3",
  "gitCommit": "434bfd82814af038ad94d62ebe59b133fcb50506",
  "gitTreeState": "clean",
  "buildDate": "2022-10-25T19:35:11Z",
  "goVersion": "go1.19.2",
  "compiler": "gc",
  "platform": "linux/amd64"
}
```

This would be required for those Kubernetes distributions which are protecting the version endpoint, thus a valid token, and a user along with groups, must be enforced.

It would be great if you could give it a try, ensuring it works as expected in your environment.

sybnex:
Hi, 

thanks for your big effort! The part, you developed, with the version works as designed!

Unfortunately it didn't solved the problem. I increased all debug logs to the max but couldn't find any hint what's the problem with the dashboard. It is still not starting and hangs in a CrashLoopBackOff. 

I even tried to create an own tenant for the dashboard with his serviceaccount as owner but wasn't successfully.
Last thing would be sniffing the network ... 

Is there any change to increase the logs of the capsule-proxy by more than 10?

prometherion:
Unless sharing more details, such the reason for the crashloopbackoff, dashboard logs, and further details, I cannot help you.

I would suggest moving the discussion to the issue itself, since this place must be used mostly to address changes and technical discussions.

oliverbaehler:
Stale, Reopen on activity

PR #362, opened 2023-12-27
feat: add pod annotations to deployment manifest

abhinandanbaheti:
This is to allow setting pod annotations to deployment manifests


prometherion:
Any reason for getting this closed, @abhinandanbaheti?

I think it could be a good addition to the Helm Chart.

PR #306, opened 2023-06-21
fix: retrieving groups from header values

prometherion:
Closes #305.

@msergg may I ask you to give a try to these changes to your environment? Unfortunately, I don't have an Active Directory backed environment: you need to build on your own and push to a container registry.

https://github.com/clastix/capsule-proxy/blob/193bdab39373d0dfa470ae09956dbb38f25bb069/Makefile#L20-L22

PR #375, opened 2024-01-23
feat(controller): add impersonation group filter options

oliverbaehler:
Adds options to filter groups that will be impersonated. We experienced based on the correction in (https://github.com/projectcapsule/capsule-proxy/pull/331) that the performance greatly decreased because of the amount of groups we have in the claim and the resulting requests to the API server.

With these options the amount of groups which are considered for impersonated can be decreased resulting in less requests.

@MaxFedotov wdyt? Do you also experience performance issues? did you also change the qos and burst of the kube client?

Maybe we can do additional adjustments to improve the performance. But i can't think of anything else to decrease the requests made to the API to request impersonation.  Batching all the groups would help, however there's nothing on the kubernetes API which would offer such a function.

Maybe adding FlowSchemes and APIPriority might also help.

### DIFF
--- a/internal/options/kube.go
+++ b/internal/options/kube.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"regexp"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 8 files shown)


oliverbaehler:
@prometherion Yes sure

prometherion (approved):
LGTM, I remember we had similar issues with #306, maybe @msergg and @y0psolo could be interested.

PR #389, opened 2024-02-18
feat: feature gate for ProxyAllNamespaced

### LINKED ISSUE #185: Intercept any API/Kind
We talked to some of our developers and they ask why they can't execute commands like `kubectl get task -A`(Tekton Task CR) an just get all the tasks they have access to with that command. 

And i thought it would be a great feature to dynamically configure which resources are intercepted for which users. For example something like this:

```
kind: Tenant
metadata:
  name: oil
spec:
  owners:
  - kind: User
    name: alice
    # ...
    proxySettings:
    - api: tekton.dev/v1beta1
      kind: Task
      operations:
      - List
    - api: tekton.dev/v1beta1
      kind: Pipeline
      operations:
      - List
    - kind: PriorityClasses
      operations:
      - List
```

So as little as I understand from the code it would be difficult to change the implementation for the currently proxied resources/modules. However that's why I added an `api` field so we could differ between supported modules and a generic api and kind. Those we would have to handle differently. 

Before trying to create a solution I wanted to hear your opinions. For us this would help improve the customer experience tremendously and we need something like that in place.

prometherion:
Closes #185 as a rework of #308 that can be discarded.


oliverbaehler (approved):
Awesome work!

PR #421, opened 2024-04-10
feat(controller): add generic cluster scoped resources to proxysettings

oliverbaehler:
With this feature you can grant `LIST` permissions to cluster-scoped resources based on Labels. This Feature is added as FeatureGate and therefor not active on default installations of the capsule-proxy. Usage with the Flag `--feature-gates=ProxyClusterScoped=true`
## Example 

Here's an example implementation of the clusterResource settings:

```

---
kind: Tenant
metadata:
  name: solar
spec:
  owners:
  - name: alice
    kind: User

```

Create a namespace as alice to propagate proxySettings:

```
kubectl --kubeconfig hack/alice.kubeconfig create ns solar-test
```

Create the ProxySettings:

```
---
kind: ProxySetting
metadata:
  name: solar
  namespace: solar-test
spec:
  subjects:
  - name: alice
    kind: User
    clusterResources:
    - apiGroups:
      - kyverno.io/*
      operations:
      - "List"
      resources:
      - "*"
      selector:
        matchLabels:
          consumer: customer
    - apiGroups:
      - admissionregistration.k8s.io/v1
      resources:
      - "mutatingwebhookconfigurations"
      selector:
        matchLabels:
          consumer: tenant
    - apiGroups:
      - admissionregistration.k8s.io/v1
      resources:
      - "validatingwebhookconfigurations"
```

This setting allows alice to:

- `LIST` all `*. kyverno.io/*` resources  with the label-value  `consumer: customer`
- `LIST` all `mutatingwebhookconfigurations.admissionregistration.k8s.io/v1` with the label-value `consumer: tenant`

Label some resources that they match and run some commands:

```
$ kubectl --kubeconfig hack/bob.kubeconfig get mutatingwebhookconfigurations.admissionregistration.k8s.io
No resources found

$ kubectl --kubeconfig hack/alice.kubeconfig get mutatingwebhookconfigurations.admissionregistration.k8s.io
NAME                                     WEBHOOKS   AGE
capsule-mutating-webhook-configuration   4          84m

 kubectl --kubeconfig hack/alice.kubeconfig get clusterPolicy
NAME                    ADMISSION   BACKGROUND   VALIDATE ACTION   READY   AGE   MESSAGE
disallow-host-process   true        true         Audit             True    46m   Ready

(No Label on any resource)
$ kubectl --kubeconfig hack/alice.kubeconfig get clustercleanuppolicies.kyverno.io
No resources found
```

## Helm CRD lifecycle

This version also introduces a new flow how we can manage the CRDs with Helm beyond the install lifecycle. It's a simple template which renders the crds and by default keeps the crds even if the release is deleted. In the Chart README.md the entire upgrading process is documented.

This change is by default not breaking, because the current lifecycle process is not active. This way our users can opt-in and perform the required actions for the upgrade (documented in the README.md of the chart). All these changes are done with this commit:

* https://github.com/projectcapsule/capsule-proxy/pull/421/commits/d98c60bf3defbe0032b7ad211e6bafe9fd203127

When deleting the chart, the default behavoir is to keep the installed CRDs:

```
These resources were kept due to the resource policy:
[CustomResourceDefinition] proxysettings.capsule.clastix.io

release "capsule-proxy" uninstalled
```

**Note**: Should this change be accepted, I would like to introduce this cycle on the capsule helm chart as well, but with the next major release.


prometherion (approved):
The feature is very neat and I like it, thanks for the big efforts you're pushing in leading Capsule an the Proxy features.

I'm a bit frightened of the CRD handling, however, and CI is complaining: wondering if it's related to it.

If we can assess everything's working good, +1 from me.

PR #422, opened 2024-04-11
feat(controller): add --skip-impersonation-review option

### LINKED ISSUE #420: feat: option to skip impersonation review for proxy environments (teleport)
# Describe the feature

When using impersonation, the proxy will create for each group an `SubjectAccessReview`. We can't batch this operation and therefor the more groups you are sending as Impersonation headers, the more requests are hitting the API. We have noticed, that this can make things slow (when people are working parallel some queries take like 20s). I have already proposed other features which help improve the situations by filtering out the groups which are necessary for impersonation and dropping others.

- https://github.com/projectcapsule/capsule-proxy/pull/375

However in this scenario we had, IMHO we could have just skipped that process. Take for reference how Teleport connects a cluster:

https://goteleport.com/docs/kubernetes-access/introduction/

What's essentially happening:

1. The Teleport Proxy sends all the user meta to the cluster agent (Teleport Kubernetes Service)
2. The Teleport Kubernetes Service has a ServiceAccount which can impersonate anything. In that case the origin Username and the groups given via Teleport policies are Impersonated from the Teleport Kubernetes Service.
3. Because the ServiceAccount can impersonate anything, any SubjectAccessReview will always be successful. And here arises my point, since the entire Teleport Stack already solves Authentication/Authorization the capsule-proxy just slows down the entire process. The capsule-proxy only needs to filter the requests essentially.

Add a new flag `--skip-impersonation-review` which is by default `false`. When setting `true` the impersonation properties for all the requests (user and groups) are not reviewed by the proxy through the Kubernetes API.

oliverbaehler:
With this feature you can disable the capsule proxy to review the impersonated properties. Scenarios where this might be relevant are describe in the issue: https://github.com/projectcapsule/capsule-proxy/issues/420

It also fixes the other properties for the impersonation which were added with:

* https://github.com/projectcapsule/capsule-proxy/pull/375

They should be also functional, no clue where these changes got lost :D 

Closes #420 🍀


main.go
```diff
@@ -83,6 +83,7 @@ func main() {
 	flag.StringSliceVar(&ignoredUserGroups, "ignored-user-group", []string{}, "Names of the groups which requests must be ignored and proxy-passed to the upstream server")
 	flag.StringSliceVar(&ignoreImpersonationGroups, "ignored-impersonation-group", []string{}, "Names of the groups which are not used for impersonation (considered after impersonation-group-regexp)")
 	flag.StringVar(&impersonationGroupsRegexp, "impersonation-group-regexp", "", "Regular expression to match the groups which are considered for impersonation")
+	flag.BoolVar(&skipImpersonationReview, "skip-impersonation-review", false, "Skip verifiying the impersonation user and groups with subjectaccessreviews")
```
prometherion, line 86: Let's describe clearly the implications in disabling the impersonation review, with some all caps.

prometherion (changes requested):
Also in the main, let's highlight if the option is enabled and all the bad things that could happen with this option enabled: you must be aware of what you're doing.

Furthermore, don't you think we could introduce this as a FeatureGate? 🤔

maxgio92:
It would make sense @prometherion.
That would exponentially reduce the probability to disable the review by mistake

prometherion (approved):

prometherion:
I think we can merge this, it's a +1 for me.

We need also efforts to document this on the Capsule documentation regarding the Proxy.

maxgio92:
LGMT

maxgio92 (approved):

PR #486, opened 2024-07-22
feat(chart): additional deployment object configurations

oliverbaehler:
Additional Values


prometherion (commented):
Shouldn't we rename the commit and PR too with additional Deployment object configurations? We're going to support probes, volumes, env, etc.

prometherion (approved):

PR #532, opened 2024-09-27
feat(chart): add autoscaling values

oliverbaehler:
adds some quality of life improvements to the helm chart

### DIFF
--- a/charts/capsule-proxy/values.yaml
+++ b/charts/capsule-proxy/values.yaml
@@ -9,6 +9,13 @@ crds:
   # -- Keep the CustomResourceDefinitions (when the chart is deleted)
   keep: true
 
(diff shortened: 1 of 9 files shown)


charts/capsule-proxy/values.yaml
```diff
@@ -9,6 +9,13 @@ crds:
   # -- Keep the CustomResourceDefinitions (when the chart is deleted)
   keep: true
 
+# Manage RBAC Lifecycle
+rbac:
+  # -- Enable Creation of ClusterRoles
+  enabled: true
```
prometherion, line 15: Shouldn't it be `false` by default, since `cluster-admin` role already exists?

prometherion (approved):

Issue #499, opened 2024-08-07
add support for Websockets


CrimsonFez:
# Describe the feature

Add support for reverse proxying websockets.

# Expected behavior

When a client tries to open a websocket for something like exec or log, it should work.

prometherion:
The Capsule Proxy already supports websocket connections:

```
$: KUBECONFIG=capsule-proxy.kubeconfig kubectl get pods -A
NAMESPACE      NAME                                   READY   STATUS    RESTARTS   AGE
qwe12345-gpu   363b1077-5d99-45bb-abb1-eb3d17d09bf2   1/1     Running   0          42s

$: KUBECONFIG=capsule-proxy.kubeconfig kubectl -n qwe12345-gpu logs -f 363b1077-5d99-45bb-abb1-eb3d17d09bf2/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: info: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
/docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2024/08/07 06:38:07 [notice] 1#1: using the "epoll" event method
2024/08/07 06:38:07 [notice] 1#1: nginx/1.27.0
2024/08/07 06:38:07 [notice] 1#1: built by gcc 12.2.0 (Debian 12.2.0-14) 
2024/08/07 06:38:07 [notice] 1#1: OS: Linux 5.15.0-113-generic
2024/08/07 06:38:07 [notice] 1#1: getrlimit(RLIMIT_NOFILE): 1048576:1048576
2024/08/07 06:38:07 [notice] 1#1: start worker processes
2024/08/07 06:38:07 [notice] 1#1: start worker process 29
[...]

PR #533, opened 2024-09-30
feat: add support for bearer tokens in websocket protocols

CrimsonFez:
This is a simple change that enables us to extract the JWT from the WebSocket protocols as defined here https://github.com/kubernetes/kubernetes/pull/47740. This satisfies #499.

Let me know if any changes are needed.

Thank you!

### DIFF
--- a/internal/request/http.go
+++ b/internal/request/http.go
@@ -158,24 +158,24 @@ func (h http) processBearerToken() (username string, groups []string, err error)
 func (h http) bearerToken() (string, error) {
 	tradBearer := strings.ReplaceAll(h.Header.Get("Authorization"), "Bearer ", "")
 	wsHeader := h.Header.Get("Sec-Websocket-Protocol")
-	fmt.Println("wsheader:", wsHeader)
+
 	switch {
 	case tradBearer != "":
 		return tradBearer, nil
 	case wsHeader != "":
-		re := regexp.MustCompile(`(base64url\.bearer\.authorization\.k8s\.io\.)([^,]*)`)
-		match := re.FindStringSubmatch(wsHeader)
-		if len(match) < 3 {
-			return "", NewErrUnauthorized("failed to extract token from Websocket Protocols")
-		}
-		if match[2] != "" {
+		re := regexp.MustCompile(`base64url\.bearer\.authorization\.k8s\.io\.([^,]*)`)
+
+		match := re.FindStringSubmatch(wsHeader)[1]
+		if match != "" {
 			// our token is base64 encoded without padding
-			b64decode, err := base64.RawStdEncoding.DecodeString(match[2])
+			b64decode, err := base64.RawStdEncoding.DecodeString(match)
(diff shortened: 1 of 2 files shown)


oliverbaehler:
HI! You need to sign off your commits and gpg sign them as well :)

CrimsonFez:
Should now be correct

internal/request/http.go
```diff
@@ -148,8 +149,29 @@ func (h http) processBearerToken() (username string, groups []string, err error)
...
+			fmt.Println("failed to decode websocket auth bearer:", err)
+		}
+		if match[2] != "" {
+			return string(b64decode)
+		} else {
+			return ""
+		}
+	} else {
+		return ""
+	}
```
prometherion, line 174:
```suggestion
	switch {
	case tradBearer != "":
		return tradBearer
	case wsHeader != "":
		re := regexp.MustCompile(`(base64url\.bearer\.authorization\.k8s\.io\.)([^,]*)`)
		match := re.FindStringSubmatch(wsHeader)
		if len(match) <= 3 {
			panic("token for was doesn't match regex")
		}
		// our token is base64 encoded without padding
		b64decode, err := base64.RawStdEncoding.DecodeString(match[2])
		if err != nil {
			panic(fmt.Sprintf("failed to decode websocket auth bearer: %s", err.Error()))
		}
		if match[2] == "" {
			panic("missing token from ws")
		}

		return string(b64decode)
	default:
		panic("cannot retrieve token from Bearer or Sec-Websocket-Protocol headers")
	}
```

prometherion (changes requested):
Maybe the suggested code is incorrect, but we cannot return an empty token, otherwise the following calls would inherit the capsule-proxy one leading to a potential privilege escalation.

CrimsonFez:
You have a good point. 
However, I believe that later parts of the code check if the token string is empty and return an error stating that we cannot authenticate anonymous users.

prometherion:
@CrimsonFez you're right, touché!

Given the increased complexity it would be good if we could return an error and make it visible in the logs: we can achieve that with `panic` which is not ideal per se, or rather, by returning a second return such as the `error`.

prometherion (changes requested):
LGTM, happy to get this merged, just ensure the changes comply with the linters:

```
make golint
```

prometherion (approved):
Super happy getting this merged, thanks a lot! 🚀

prometherion:
Ah, `golangci-lint` complains about minor style stuff: this should be the latest round, please @CrimsonFez, may I ask you to amend it?

You can test with `golangci-lint` by issuing `make golint`

CrimsonFez:
Sure thing. I thought I got it all ;(

CrimsonFez:
Thanks for cleaning that up! I don't always have a lot of time to do stuff. Very much appreciated!

prometherion:
It's team work, thanks, mate!

PR #566, opened 2024-11-07
fix: support for service account bound tokens expiration

ppodevlabs:
Fixes #565 

From kubernetes 1.30 kubelet takes care of renewing tokens for service accounts inside the projected volumes. This PR make capsule-proxy to reload the token when required.

PR #569, opened 2024-11-11
fix: support service account bound token expiration

ppodevlabs:
Fixes #566 

From kubernetes 1.30 kubelet takes care of renewing tokens for service accounts inside the projected volumes. This PR make capsule-proxy to reload the token when required.

### DIFF
--- a/internal/webserver/webserver.go
+++ b/internal/webserver/webserver.go
@@ -11,11 +11,13 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/textproto"
+	"os"
 	"regexp"
 	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/golang-jwt/jwt/v5"
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


internal/webserver/webserver.go
```diff
@@ -97,6 +101,8 @@ type kubeFilter struct {
 	skipImpersonationReview    bool
 	reverseProxy               *httputil.ReverseProxy
 	bearerToken                string
+	bearerTokenFile            string
+	bearerTokenExpirationTime  int64
```
prometherion, line 105: couldn't we easily store `time.Time`, rather than an `int64`?

ppodevlabs: fixed

internal/webserver/webserver.go
```diff
@@ -502,3 +508,29 @@ func (n *kubeFilter) removingHopByHopHeaders(request *http.Request) {
 
 	request.Header.Del(connectionHeaderName)
 }
+
+func (n *kubeFilter) BearerToken() string {
+	if !n.bearerTokenExpirationTime.After(time.Now()) {
```
prometherion, line 513:
```suggestion
	if time.Now().After(!n.bearerTokenExpirationTime) {
```

Isn't this more readable? 🤔

prometherion (approved):

PR #681, opened 2025-03-27
fix: prevent unauthorized deletion of cluster-scoped objects

simonfuhrer:


oliverbaehler (approved):
Thanks for spotting that 🏅

oliverbaehler (approved):

PR #746, opened 2025-06-03
chore: set loglevel of token to debug

sandert-k8s:
Set loglevel of the token to debug (10) so it doesn't get shown in stdout log by default (4)


sandert-k8s:
DCO check fails, but i have signed off 🤷🏼‍♂️

prometherion (approved):

oliverbaehler (approved):

PR #891, opened 2025-09-19
feat(charts/capsule-proxy): added extra manifests in values file

Llyth:
This will give the possibility to add extra kubernetes resources/objects such as (Global)ProxySettings, KEDA ScaledObject, configMaps.
Things that are not part of the chart itself but part of/related to Capsule proxy.


Llyth:
cf. https://github.com/projectcapsule/capsule/pull/1653#issuecomment-3346153847

Made some slight modification to stay compliant with CI

Llyth:
Following the identical PR in Capsule: https://github.com/projectcapsule/capsule/pull/1653

And following comment: https://github.com/projectcapsule/capsule/pull/1653#issuecomment-3371328250

I've modified the PR to instead use a random configMap in the `extraManifests`.

_In my case I do use it for GlobalProxySettings, but as to not trigger any conflict with a webhook not being ready **in time** during install, the example is modified._

Successfully ran in local:
- `make helm-docs`
- `make helm-schema`
- `make helm-test`

Best regards.

PR #1041, opened 2026-05-29
feat: add observedgeneration to status object of all crds

sandert-k8s:
Pre work for https://github.com/projectcapsule/capsule/issues/1924

Add observedGeneration to the status of all objects. ArgoCD best practices is to check observedGeneration vs Generation in a spec. If that differs, the application will report "Progressing".
Also added controllers for (global)ProxySettings, since they didn't have them yet.


internal/controllers/observed_generation.go
```diff
@@ -0,0 +1,71 @@
...
+					req.NamespacedName, err,
+				)
+			}
+
+			return err
+		}
+
+		return nil
+	})
+}
```
oliverbaehler: While this is very pragamtic i would prefer keeping the same pattern as in the maincontroller:

1. Getting instance of current object
2. Post Reconciling Condition
3. Post Status (Defer)

internal/controllers/observed_generation.go
```diff
@@ -0,0 +1,71 @@
...
+			}
+
+			return err
+		}
+
+		if getObservedGeneration(latest) == latest.GetGeneration() {
+			return nil
+		}
+
+		original, ok := latest.DeepCopyObject().(client.Object)
```
oliverbaehler: DeepCopy is always very expensive in terms of performance

oliverbaehler (changes requested):

internal/controllers/observed_generation.go
sandert-k8s: Yup good point, fixed so it matches the logic of Capsule!

internal/controllers/observed_generation.go
sandert-k8s: Check, removed DeepCopy

oliverbaehler (approved):

PR #1043, opened 2026-05-30
fix: return early after error in impersonateHandler

jouve:
When the user or group lookup fails, the handler was logging/responding the error but then falling through to the impersonation logic. Adding an early return prevents the request from continuing with an empty or invalid identity.


oliverbaehler (approved):
0

PR #1069, opened 2026-07-07
feat: improve SSAR and SSRR performance and usability

lukasboettcher:
This PR lets `authorizationMiddleware` requests pass through to the upstream apiserver if bearer token auth is used. The SSAR response gets modified to include cluster-wide permissions for namespaced resources. 
Additionally this PR lets non tenant owners get namespace list permissions from GlobalProxySettings.
This has a few advantages:
- all upstream SSAR/SSRR permissions are kept, capsule-proxy only appends the permissions it additionally grants
- the response times are drastically improved for users with lots of groups / roles in their id token claims
- the SSAR response correctly display the permissions for `kubectl get pods -A`
- non-owners can be explicitly assigned cluster scoped namespace permissions without reliance on the rolebinding reflector

Reasoning: we use AKS with capsule/-proxy. In AKS clusters it is common for users to have more than 300 groups assigned in Entra ID for large tenants. If we let capsule-proxy impersonate each group of the user's token, the upstream apiserver needs to perform the permission review for each impersonated group individually.
This is extremely slow and causes timeouts in lots of k8s clients.

Could be related to https://github.com/projectcapsule/capsule-proxy/issues/989

Example for a user with ~300 groups:

before
```
$ time kubectl auth can-i --list
Resources                          Non-Resource URLs   Resource Names   Verbs
namespaces                         []                  []               [List]
nodes./v1                          []                  []               [list]
storageclasses.storage.k8s.io/v1   []                  []               [list]

real    0m14.439s
user    0m0.279s
sys     0m0.082s

$ kubectl auth can-i list pods --all-namespaces
no

$ kubectl get ns
test-1
```
after
```
$ time kubectl auth can-i --list
Resources                                       Non-Resource URLs   Resource Names   Verbs
namespaces                                      []                  []               [create patch list]
selfsubjectreviews.authentication.k8s.io        []                  []               [create]
selfsubjectaccessreviews.authorization.k8s.io   []                  []               [create]
selfsubjectrulesreviews.authorization.k8s.io    []                  []               [create]
                                                [/api/*]            []               [get]
                                                [/api]              []               [get]
                                                [/apis/*]           []               [get]
                                                [/apis]             []               [get]
                                                [/healthz]          []               [get]
                                                [/healthz]          []               [get]
                                                [/livez]            []               [get]
                                                [/livez]            []               [get]
                                                [/openapi/*]        []               [get]
                                                [/openapi]          []               [get]
                                                [/readyz]           []               [get]
                                                [/readyz]           []               [get]
                                                [/version/]         []               [get]
                                                [/version/]         []               [get]
                                                [/version]          []               [get]
                                                [/version]          []               [get]
nodes./v1                                       []                  []               [list]
storageclasses.storage.k8s.io/v1                []                  []               [list]

real    0m0.584s
user    0m0.295s
sys     0m0.079s

$ kubectl auth can-i list pods --all-namespaces
yes

$ kubectl get ns
asdf-1
test-1
```

config:
```
values:
          disableCaching: true
          extraArgs:
            - "--feature-gates=ProxyClusterScoped=true"
            - "--feature-gates=ProxyAllNamespaced=true"
---
kind: GlobalProxySettings
metadata:
  name: global-proxy-settings
spec:
  rules:
    - subjects:
        - name: uuid
          kind: Group
      clusterResources:
        - apiGroups:
            - "storage.k8s.io/v1"
          resources:
            - "storageclasses"
          operations:
            - List
          selector:
            matchExpressions:
              - key: private
                operator: DoesNotExist
        - apiGroups:
            - "/v1"
          resources:
            - "nodes"
          operations:
            - List
          selector:
            matchExpressions:
              - key: private
                operator: DoesNotExist
        - apiGroups:
            - "/v1"
          resources:
            - "namespaces"
          operations:
            - List
          selector:
            matchLabels:
              capsule.clastix.io/tenant: asdf
---
kind: Tenant
metadata:
  name: test
spec:
  owners:
    - name: uuid
      kind: Group
---
kind: Tenant
metadata:
  name: asdf
spec:
  additionalRoleBindings:
  - clusterRoleName: edit
    subjects:
    - name: uuid
      kind: Group
```

### DIFF
--- a/internal/modules/namespace/utils.go
+++ b/internal/modules/namespace/utils.go
@@ -1,0 +1,69 @@
+// Copyright 2020-2025 Project Capsule Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package namespace
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/projectcapsule/capsule-proxy/internal/modules/clusterscoped"
+	"github.com/projectcapsule/capsule-proxy/internal/modules/utils"
+	"github.com/projectcapsule/capsule-proxy/internal/tenant"
+	"github.com/projectcapsule/capsule-proxy/internal/types"
+)
+
+// namespacesGVK returns the GroupVersionKind used to match cluster-scoped
+// ClusterResources rules against the core namespaces resource.
+func namespacesGVK() *schema.GroupVersionKind {
+	return &schema.GroupVersionKind{Group: corev1.GroupName, Version: types.V1, Kind: types.Namespaces}
+}
+
+// clusterScopedNamespaceNames resolves the namespaces selected by cluster-scoped
+// ClusterResources rules (matching the namespaces resource) to their names, so
+// subjects granted access via GlobalProxySettings or ProxySettings can list the
+// corresponding namespaces without being tenant owners.
+func clusterScopedNamespaceNames(ctx context.Context, reader client.Reader, proxyTenants []*tenant.ProxyTenant) ([]string, error) {
+	_, requirements := clusterscoped.GetClusterScopeRequirements(namespacesGVK(), proxyTenants)
+	if len(requirements) == 0 {
+		return nil, nil
+	}
+
+	selector, err := utils.HandleListSelector(requirements)
+	if err != nil {
+		return nil, err
+	}
+
+	nsList := &corev1.NamespaceList{}
+	if err := reader.List(ctx, nsList, client.MatchingLabelsSelector{Selector: selector}); err != nil {
+		return nil, err
+	}
+
+	names := make([]string, 0, len(nsList.Items))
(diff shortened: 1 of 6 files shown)


internal/modules/namespace/utils.go
```diff
@@ -0,0 +1,69 @@
...
+
+	nsLabels := labels.Set(ns.GetLabels())
+	for _, requirement := range requirements {
+		if requirement.Matches(nsLabels) {
+			return true
+		}
+	}
+
+	return false
+}
```
oliverbaehler, line 69: Can we get some unit tests for these?

oliverbaehler (changes requested):
Very well crafted Changes/PR, thank you very much.

You added to possibility that additional namespaces can be granted via GlobalProxySettings. I like that change. However we should then also add this same functionality for Tenants itself. Since these two resources are always handeled by us, even if the ClusterScoped feature-flag is enabled. WDYT

internal/modules/namespace/utils.go
oliverbaehler: @copilot suggest unit tests

oliverbaehler (approved):

PR #1097, opened 2026-08-27
fix: empty or malformed JWTs now produce a zero expiration without dereferencing a nil token

oliverbaehler:
Reported via #1095


Svarrogh1337 (approved):
Lgtm

PR #1112, opened 2026-09-23
fix: ignore clusterResources in namespaced proxysettings

oliverbaehler:
Tenant-writable ProxySetting objects could grant cluster-scoped reads through the proxy’s privileged credentials, breaking tenant isolation. Ignore namespaced clusterResources at runtime, including previously stored grants, and reject new nonempty grants through CRD validation. Preserve tenant-local delegation, legacy proxy operations, and administrator-managed GlobalProxySettings.


internal/tenant/proxytenant.go
```diff
@@ -37,26 +37,21 @@ func NewProxyTenant(
...
 	}
 
-	pt := &ProxyTenant{
-		Tenant:           tenant,
-		ClusterResources: tenantClusterResources,
-	}
+	// Namespaced ProxySettings are tenant-writable and cannot grant cluster-wide
+	// access. Ignore ClusterResources even on objects stored before CRD validation
+	// was tightened. Only NewClusterProxy may import those administrator grants.
+	pt := &ProxyTenant{Tenant: tenant}
```
bakito: should it produce a warn log or an event if such a config is applied

Svarrogh1337 (approved):

